PR #70, opened 2015-01-20
This should fix a path enumeration issue

dxa4481:
I got rid of normilize [here](https://github.com/dxa4481/send/blob/master/index.js#L420) because join normilizes, and I added [this](https://github.com/dxa4481/send/blob/master/index.js#L423) to resolve the issue found here https://github.com/TeamMentor/TM_4_0_Design/issues/206

### DIFF
--- a/index.js
+++ b/index.js
@@ -421,7 +421,7 @@ SendStream.prototype.pipe = function(res){
     root = normalize(root + sep)
 
     // check for exiting current dir
-    var traversalCheck = rawPath.split(sep);
+    var traversalCheck = rawPath.split('/');
     var topLevel = '';
     for (var i = 0; i < traversalCheck.length; i++) {
       topLevel += traversalCheck[i];
--- a/index.js
+++ b/index.js
@@ -416,7 +416,8 @@ SendStream.prototype.pipe = function(res){
   var parts
   if (root !== null) {
     // join / normalize from optional root dir
-    path = normalize(join(root, path))
+    rawPath = root + path
+    path = join(root, path)
     root = normalize(root + sep)
 
     // malicious traversal
--- a/index.js
+++ b/index.js
@@ -419,6 +419,12 @@ SendStream.prototype.pipe = function(res){
     path = normalize(join(root, path))
     root = normalize(root + sep)
 
+    // malicious traversal
+    if(rawPath !== path){
+      debug('malicious path "%s"', path)
+      return this.error(403)
+    }
+
     // malicious path
     if ((path + sep).substr(0, root.length) !== root) {
       debug('malicious path "%s"', path)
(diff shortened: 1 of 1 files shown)


dougwilson:
Great! I can look into why the changes are making the tests fail.

dougwilson:
The change also seems to make pretty much every single test fail on Windows.

dougwilson:
> The change also seems to make pretty much every single test fail on Windows.

Looks like the main reason for the Windows failures is because `join(root, path)` will never equal `root + path` under any circumstance because of the fact that URLs have forward slashes and Windows uses backslashes in file paths.

dxa4481:
Ahh, gotcha. Sorry I didn't run the tests; I set up the test suite now and am having another go at it.

dougwilson:
No problem :) I just haven't yet gotten to a machine to be able to look into anything yet. Your effort is greatly appreciated!

dougwilson:
Oh, and I think if I were to give the fix a goal, it would be that "at no point, should the path ever descend below the `root`".

dougwilson:
Ok, I got back and I have a perfect solution :)! Basically, just normalize the rhs first, then there is a simple check for a leading up directive.

dxa4481:
Oh man! That's probably simpler than what I ended up doing, but the tests pass with my latest commit

index.js
```diff
@@ -416,9 +416,22 @@ SendStream.prototype.pipe = function(res){
   var parts
   if (root !== null) {
     // join / normalize from optional root dir
-    path = normalize(join(root, path))
+    var rawPath = path
+    path = join(root, path)
     root = normalize(root + sep)
 
+    // check for exiting current dir
+    var traversalCheck = rawPath.split(sep);
```
dougwilson: This still doesn't work on Windows ;) `sep === '\\'` on there, and `path` will just contain forward slashes.

dougwilson: I think _eventually in the far distant future_ we'll finally have `path.split` in Node.js core, haha.

dougwilson:
Ok, @dxa4481 @DinisCruz if you guys want to confirm the current `master` (`npm install tj/send`) and poke at it, I would love a confirmation :) Then I'll roll it up into Express 4.11.1 and 3.19.1, and whatever else I can get my hands on.

DinisCruz:
Great stuff, I'll take a look and its great to see such a fast fix.

I will try to also add a unit test that checks for the edge cases discovered

dougwilson:
Awesome, @DinisCruz :) The final commit does have _one_ unit test (https://github.com/tj/send/commit/98a5b89982b38e79db684177cf94730ce7fc7aed), but if you want to add more, that would be great :)

dougwilson:
So far `send` has been published as 0.11.1

PR #78, opened 2015-06-25
Add `fs` option for using other filesystem implementations.

izaakschroeder:
Options can now take an `fs` parameter which must emulate the functionality of `fs.stat` and `fs.createReadStream`.

The lovely webpack has `memory-fs` which (mostly) implements the required functions; but then they go and do something silly like rewrite a static file sender for `webpack-dev-middleware` when there's a perfectly good one here. This will enable the transport of files from `memory-fs` as if they were local.

See also https://github.com/webpack/memory-fs/pull/5

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,8 +87,28 @@ Serve files relative to `path`.
 
 Use a filesystem other than the node-js built-in `fs` module. It must support the following methods:
 
- * stat(path)
- * createReadStream(path)
+ * stat(path, callback)
+ * createReadStream(path, options)
+
+The resulting stat object from `stat` must (at least) look like:
+
+```javascript
+{
+    mtime: Date,
+    ctime: Date,
+    ino: Number,
+    size: Number
+}
+```
+
+The resulting readable stream must (at least) support the options:
+
+```javascript
+{
+    start: Number,
+    end: Number
+}
+```
 
 ### Events
 
--- a/README.md
+++ b/README.md
@@ -87,28 +87,47 @@ Serve files relative to `path`.
 
 Use a filesystem other than the node-js built-in `fs` module. It must support the following methods:
 
- * stat(path, callback)
- * createReadStream(path, options)
+ * `stat(path, callback)`
+ * `createReadStream(path, options)``
 
-The resulting stat object from `stat` must (at least) look like:
+The `callback` given to `stat` has two parameters: `(err, result)`.
 
-```javascript
-{
-    mtime: Date,
-    ctime: Date,
-    ino: Number,
-    size: Number
-}
-```
-
-The resulting readable stream must (at least) support the options:
+`err`: The `err` parameter should be an object with a `code` property containing error information.
 
 ```javascript
 {
-    start: Number,
-    end: Number
+    code: String // Error code.
 }
 ```
+
+The following codes are respected: 'ENAMETOOLONG', 'ENOENT', and 'ENOTDIR'. Descriptions of these errors can be found [here](https://github.com/rvagg/node-errno/blob/master/errno.js).
+
+`result`: The `result` stat object from `stat` must (at least) look like:
+
+```javascript
+{
+    mtime: Date, // Modification date
+    ctime: Date, // Creation date
+    ino: Number, // Unique id
+    size: Number, // Length of object, in bytes
+    isDirectory: Function // Returns true if object is directory
+}
+```
+
+NOTE: The `ino` property represents the file-system inode, which is unique to the given _filesystem_ meaning that for `fs` where it's possible to access multiple mounted filesystems `ino` is not unique, but for any virtual driver it would be reasonable to ensure that `ino` numbers are.
+
+For complete details about these (and other) properties on the `stat` object be sure to see the [documentation](https://nodejs.org/api/fs.html#fs_class_fs_stats).
+
+The `createReadStream` method must (at least) support the options:
+
+```javascript
+{
+    start: Number, // Start offset of data to read
+    end: Number // End offset of data to read
+}
+```
+
+For complete details about these (and other) options of the `createReadStream` method be sure to see the [documentation](https://nodejs.org/api/fs.html#fs_fs_createreadstream_path_options).
 
 ### Events
 
--- a/README.md
+++ b/README.md
@@ -83,6 +83,13 @@ This can also be a string accepted by the
 
 Serve files relative to `path`.
 
+##### fs
+
+Use a filesystem other than the node-js built-in `fs` module. It must support the following methods:
+
+ * stat(path)
+ * createReadStream(path)
+
 ### Events
 
 The `SendStream` is an event emitter and will emit the following events:
@@ -177,7 +184,7 @@ var app = http.createServer(function(req, res){
 }).listen(3000);
 ```
 
-## License 
+## License
 
 [MIT](LICENSE)
 
--- a/test/send.js
+++ b/test/send.js
@@ -7,7 +7,7 @@ var fs = require('fs');
 var http = require('http');
 var path = require('path');
 var request = require('supertest');
-var stream = require('stream');
+var ReadableStream = require('readable-stream');
 var send = require('..')
 
 // test server
@@ -1298,7 +1298,7 @@ describe('send(file, options)', function(){
               },
               createReadStream: function() {
                   createReadStreamCalled = true
-                  var bam = stream.Readable()
+                  var bam = new ReadableStream()
                   bam.push(res)
                   bam.push(null)
                   return bam;
--- a/test/send.js
+++ b/test/send.js
@@ -7,6 +7,7 @@ var fs = require('fs');
 var http = require('http');
 var path = require('path');
 var request = require('supertest');
+var stream = require('stream');
 var send = require('..')
 
 // test server
@@ -410,14 +411,14 @@ describe('send(file).pipe(res)', function(){
       .set('Range', 'bytes=0-4')
       .expect(206, '12345', done);
     })
-    
+
     it('should be inclusive', function(done){
       request(app)
       .get('/nums')
       .set('Range', 'bytes=0-0')
       .expect(206, '1', done);
     })
-    
+
     it('should set Content-Range', function(done){
       request(app)
       .get('/nums')
@@ -432,7 +433,7 @@ describe('send(file).pipe(res)', function(){
       .set('Range', 'bytes=-3')
       .expect(206, '789', done);
     })
-    
+
     it('should support n-', function(done){
       request(app)
       .get('/nums')
@@ -1063,7 +1064,7 @@ describe('send(file, options)', function(){
         send(req, 'test/fixtures/name.txt', {maxAge: Infinity})
         .pipe(res);
       });
-      
+
       request(app)
       .get('/name.txt')
       .expect('Cache-Control', 'public, max-age=31536000', done)
@@ -1275,6 +1276,48 @@ describe('send(file, options)', function(){
         .expect(200, '...', done);
       })
     })
+  })
+
+  describe('fs', function() {
+      it('should invoke correct fs methods', function(done) {
+          var statCalled = false
+          var createReadStreamCalled = false
+          var res = new Buffer('hello')
+          var fs = {
+              stat: function(file, cb) {
+                  statCalled = true
+                  cb(null, {
+                      size: res.length,
+                      ctime: new Date(),
+                      mtime: new Date(),
+                      ino: 0,
+                      isDirectory: function() {
+                          return false
+                      }
+                  })
+              },
+              createReadStream: function() {
+                  createReadStreamCalled = true
+                  var bam = stream.Readable()
+                  bam.push(res)
+                  bam.push(null)
+                  return bam;
+              }
+          }
+          var app = http.createServer(function(req, res){
+            send(req, req.url, {fs: fs})
+            .pipe(res);
+          })
+
+          request(app)
+          .get('/some/dir')
+          .expect(200, 'hello', function(err) {
+              if (err) return done(err)
+              assert.ok(statCalled)
+              assert.ok(createReadStreamCalled)
+              done()
+          })
+      })
   })
 })
 
--- a/test/send.js
+++ b/test/send.js
@@ -1280,13 +1280,14 @@ describe('send(file, options)', function(){
 
   describe('fs', function() {
       it('should invoke correct fs methods', function(done) {
+          var cb = after(3, done);
           var statCalled = false
           var createReadStreamCalled = false
           var res = new Buffer('hello')
           var fs = {
-              stat: function(file, cb) {
-                  statCalled = true
-                  cb(null, {
+              stat: function(file, next) {
+                  cb();
+                  next(null, {
                       size: res.length,
                       ctime: new Date(),
                       mtime: new Date(),
@@ -1297,7 +1298,7 @@ describe('send(file, options)', function(){
                   })
               },
               createReadStream: function() {
-                  createReadStreamCalled = true
+                  cb();
                   var bam = new ReadableStream()
                   bam.push(res)
                   bam.push(null)
@@ -1313,9 +1314,7 @@ describe('send(file, options)', function(){
           .get('/some/dir')
           .expect(200, 'hello', function(err) {
               if (err) return done(err)
-              assert.ok(statCalled)
-              assert.ok(createReadStreamCalled)
-              done()
+              cb();
           })
       })
   })
--- a/index.js
+++ b/index.js
@@ -95,6 +95,10 @@ function SendStream(req, path, options) {
     ? opts.dotfiles
     : 'ignore'
 
+  this._fs = opts.fs !== undefined
+    ? opts.fs
+    : fs;
+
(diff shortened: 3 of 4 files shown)


scottcorgan:
Yes, this!

I had to write a new package to do just this.

Would love to see something like this make it into this module.

README.md
```diff
@@ -83,6 +83,33 @@ This can also be a string accepted by the
...
+##### fs
+
+Use a filesystem other than the node-js built-in `fs` module. It must support the following methods:
+
+ * stat(path, callback)
+ * createReadStream(path, options)
+
+The resulting stat object from `stat` must (at least) look like:
+
+```javascript
```
dougwilson: Missing `isDirectory` from the list

README.md
```diff
@@ -83,6 +83,33 @@ This can also be a string accepted by the
...
+{
+    mtime: Date,
+    ctime: Date,
+    ino: Number,
+    size: Number
+}
+```
+
+The resulting readable stream must (at least) support the options:
```
dougwilson: To me, if this is going to be stated, we need to explain what these options mean and how someone can actually implement them. Otherwise, it's going to lead to issues asking about it.

README.md
```diff
@@ -83,6 +83,33 @@ This can also be a string accepted by the
...
 Serve files relative to `path`.
 
+##### fs
+
+Use a filesystem other than the node-js built-in `fs` module. It must support the following methods:
+
+ * stat(path, callback)
+ * createReadStream(path, options)
+
+The resulting stat object from `stat` must (at least) look like:
```
dougwilson: The types of `err` objects also need to be documented, for instance, we look for `ENOENT` in the `code` property and do special things, so people implementing a fake `stat` would need to know this.

test/send.js
```diff
@@ -1276,6 +1277,48 @@ describe('send(file, options)', function(){
       })
     })
   })
+
+  describe('fs', function() {
+      it('should invoke correct fs methods', function(done) {
+          var statCalled = false
```
dougwilson: Please track the calls using `after` instead of using booleans.

izaakschroeder:
@dougwilson: Updated. Let me know what you think.

izaakschroeder:
A similar feature for `serve-index` would probably make some people happy too.

dougwilson:
I'll take a look in a bit. I know there are a lot of people happily using `proxyquire` with these modules, so not sure how that fits into what you are trying to accomplish, just as a note.

izaakschroeder:
Neat! `proxyquire` overrides some module loading machinery though, and from a cursory glance at the documentation doesn't really allow the stubbed and unstubbed versions to coexist happily together. You either stub it out for a module or you don't.

izaakschroeder:
Present goal is a better `webpack-dev-server` (mainly for server-side HMR, but people need client code too so there has to be a way to ship generated files from webpack). So far wanting to use `express` and its lovely friends.

dougwilson:
Right. The biggest downside to us accepting `fs` is that the `fs` is such an enormous interface. What this means is that once we accept it as an option, we either validate that every single function from Node.js code `fs` is defined on there, or otherwise if we suddenly want to add a feature that uses `fs.readFile`, it would be a breaking change for this module, since people are going to implement the minimal functions on `fs`... this needs a lot more thought on how we can accept a custom `fs` but not require us to make a major version bump just to start using another function on the `fs` object.

izaakschroeder:
Possible option: some kind of `virtual-fs` module that defines some kind of lowest-common denominator and would have its own functionality for validating `fs` type objects; `send` would have `virtual-fs` as a peer dependency. Implementors equally have `virtual-fs` as a peer dependency; `npm` doesn't allow incompatible peer dependencies among modules.

izaakschroeder:
`virtual-fs` could also provide primitive wrappers – i.e. if your fs implements `readFile` or `readFileSync` then it's always possible to have a `createReadStream` implementation based on that.

izaakschroeder:
We could also provide some primitive feature detection. `vfs.features(fs, { links: true, readable: true })`. node's built-in `fs` would then support every feature.

dougwilson:
Hm, interesting, though I don't think peer dependencies work well, because if, say, this and `serve-index` used `virtual-fs` as a peer dependency, then, since the policy is that we must always pin foreign dependencies, we would of course require an exact version of the peer dependency. This, consequence, would require people to march forward multiple modules at once, making them less likely to upgrade their modules (and means we have to release new versions of all these modules at once all the time). Thoughst?

izaakschroeder:
Thoughts:
 * There are essentially two (or more) modules at play here: one describes the `fs` interface (the peer-dependency), and the other(s) provides utility functions.
 * The interface module would essentially have no code, just a description of how the `fs` object behaves; it might make sense that this package has the same version as `node` itself? Since we're mirroring `fs`.
 * The utility classes would be normal dependencies, so being able to do `vfs.ensureFeatures()` would come from a dependency, not a peer.

dougwilson:
Sure, but one peer dependency means that module could effectively never publish an update, because we're going to pin the version, and by npm's nature, you cannot have conflicting peer dependencies, so I guess the module should try really hard to be 100% correct on the version we pin to :)

In general, I think approaching this issue with "let's just give you a mock'd `fs` as a one-off" is simply the wrong approach. Either we need to

1) re-work parts of this library to utilize DI factories to perform specific actions in a generic way that does not require that in order to document it, we don't have to define all of Node.js's inner workings (the current README changes, though not nothing, are really, really poor and no one can implement a proper mock from reading them)

-or-

2) re-work parts of this library such that you can either sub-class the `SendStream` to just use your custom methods that you can implement using a mock fs which have very defined semantics or at least just to be able to over-write those methods on the constructed instance.

izaakschroeder:
Hum. If we define any interface that deviates from `fs` it also means updating other modules – the requirement for `ino` and `ctime` come from the duck-typing in `etag`; but you're right in that the current interface to `fs` is rather nebulous and covers a fairly large surface area.

I'm somewhat against 2. since it means more work for users instead of just passing around an `fs`.

dougwilson:
Either 1 or 2 allow you to "just pass around an `fs`" because that's the issue: once we let it in and we don't throw if not every single Node.js 0.8 `fs` method is there and working, then we pidgin hole ourselves into lots of major version bumps in the future, which is not something we desire at all.

As you can see in `etag`, we are pretty strict in what we even consider a "fake stat", including requiring properties we don't even use, since we do not want to invoke a major version bump just to use them in the future. In order to accept a `fs` option here, it needs to throw if not the full Node.js 0.8 `fs` is implemented.

dougwilson:
At the very least, the README in this PR needs to fully explain how someone can implement what would be passed into the `fs` option here--linking to other places is only supplementary, as we cannot guarantee that that link will always even be correct. Someone should be able to, by only reading the README here, be able to pass something into the `fs` option that works.

README.md
```diff
@@ -83,6 +83,52 @@ This can also be a string accepted by the
...
+    size: Number, // Length of object, in bytes
+    isDirectory: Function // Returns true if object is directory
+}
+```
+
+NOTE: The `ino` property represents the file-system inode, which is unique to the given _filesystem_ meaning that for `fs` where it's possible to access multiple mounted filesystems `ino` is not unique, but for any virtual driver it would be reasonable to ensure that `ino` numbers are.
+
+For complete details about these (and other) properties on the `stat` object be sure to see the [documentation](https://nodejs.org/api/fs.html#fs_class_fs_stats).
+
+The `createReadStream` method must (at least) support the options:
```
dougwilson: What should `createReadStream` do? Should it return a value? If so, what value?

dougwilson:
P.S. I'm not trying to be hard, just so you know :) I'm just trying to make sure it's well though-out, because we are going to do a major version bump to correspond to the Express 5.0 release, but then we cannot do another major version bump for literally 1-2 years, so we are very sensitive to introducing something that brings risk of major version bumps to do simple things, is all.

dougwilson:
P.S. I was just looking though that memory-fs library, and even with your PRs, it still wouldn't work properly with this module, since the errors it provides does not have the proper `code` properties set on them for us to make intelligent decisions (as it is, a file not found using memory-fs would result in a 500 response from this library).

izaakschroeder:
Totally understandable. Thinking about the interface to files – it could also be an option to have `etag` not accept a `stats` object directly, but instead a more "well defined" object `{ modified: xxx, length: yyy }`. Then `etag` is no longer dependant on `fs` in any meaningful way. However, we could also define a small subset of `fs.Stats` for which `etag` is compatible.

My attempts at creating a minimal fs interface (minus the error codes):

* send: `fs-readable +streams +mtime`
* etag: `fs-core +mtime`

## Features:

Cross cutting. Affect properties/options.

+mtime +sync +streams

## fs-core
```
.stat(path, callback)
.statSync(path) +sync
```

```
fs.stats {
	size: Number,
	mtime: Number +mtime
}
```

## fs-readable < fs-core
```
.readFile(path, encoding, callback)
.readFileSync(path, encoding) +sync
.createReadStream(path, options) +streams
.readdir(path)
```

## fs-writable < fs-core
```
.writeFile(path, encoding, callback)
.writeFileSync(path, encoding) +sync
.createWriteStream(path, options) +streams
.unlink(path)
```

## fs-util
```
asyncFromSync() // generate Async family methods from sync ones
readStreamFromReadFile() // generate createReadStream from readFile
writeStreamFromWriteFile() // etc.
existsFromStat() // generate exists from stat
```

izaakschroeder:
Also P.S. I realized heh. :smile: More PRs there to come.

izaakschroeder:
I also realize `+streams` is not strictly necessary, since you can always polyfill with something like `readStreamFromReadFile`.

izaakschroeder:
Since it's possible to polyfill `createReadStream` the actual surface area of the API is now fairly small, at least as far as `send` is concerned. We're left with:

```
stat
readFile
readdir
```

dougwilson:
So, that's the biggest issue here: we want to solve this, but we absolutely cannot pidgin hole this module into major version bumps. For example, in your proposal above, we won't ever be able to implement a feature that uses a file's created time (as an example) for 1-2 years from now, if we define our interface to be only what we are using now without any wiggle room.

izaakschroeder:
Well we could add some wiggle room? To the point that makes sense for this implementation? The same issue exists with any interface (not just mimicking `fs`).

izaakschroeder:
Having `etag` accept some more "well-defined" input (as in a proper object with specified properties) instead of just a `fs.Stats` object poses the same issue. Once we define what that input is, we can't suddenly ask for more later.

dougwilson:
> Once we define what that input is, we can't suddenly ask for more later.

Right, which is exactly what I don't want, thus exactly why `etag` asks for `fs.Stats`.

dougwilson:
Also, please only talk about this module here; any `etag` changes really need to be discussed in that repo, unless you want this discussion to take longer, by talking about a different module in a completely different organization :)

dougwilson:
Besides, why can't webpack simply hook into the global `fs` and simply hijack a subset of file system paths as virtual? To me, that would make it suddenly support every single module on npm that uses `fs` while still letting you access the real file system and will keep you from needing to crusade through all npm just to get everyone to take `fs` as an argument.

Another question is that don't you also need to be able to specify the `path` module that we use as well? Right now, your `fs` stubs will need to support both forward and backward slashes, Windows drives, UNC paths, etc. just to work right, and your fake `fs` would need to match the same file case-sensitivity issues that whatever the `path.normalize` we are using happens to be using, no?

izaakschroeder:
Re: `etag` module – no problem :smile: 

If using some vfs implementation that's not `fs` one could just disable etags by default? The `etag` module would still then be responsible for validating whether or not it can accept the `stats` object generated.

webpack kind of can hook into global `fs` – you can alias `fs` to some internal module of your own doing; this is possible inside the code you're packaging with webpack, but I'm serving the built files outside of that.

You could go `var fs = require('fs'); fs.readFile = ...` but that seems _extremely_ brittle, and loses all manner of contextualization; I don't want override it for everything, just certain cases. It might be possible doing something like https://github.com/izaakschroeder/hook-module/blob/master/example/test.js but that is an equally scary road.

Good point about windows paths – `memory-fs` handles them in a somewhat sensible way by simply rewriting windows paths to UNIX ones. Annoying, but again could be abstracted into a general utility function that just obviates the need to worry about Windows paths.

dougwilson:
P.S. remember: this is all discussion at this point :) If I hated the idea of being able to do virtual fs and such, I would have ended up closing the issue, haha :) I want to just be sure you don't think I'm trying to kill it--I just want to make sure we end up with an awesome, useful end result :)

dougwilson:
And @scottcorgan feel free to chime in any time on thoughts, as since you commented, I assume you're interested in seeing this come to fruition :)

izaakschroeder:
Yea I'd be interested in seeing the module @scottcorgan wrote that dealt with this issue.

izaakschroeder:
Well @dougwilson we could have two separate code paths for capabilities?

Checking VFS compatibility:
```javascript
var myvfs = { ... }

// Adding new cap would require major version bump here
if (!myvfs.capabilities.readable || !myvfs.capabilities.xxx || ...) {
  throw new TypeError('Invalid fs');
}

// Adding new cap would NOT require major version bump
if (myvfs.capabilities.mtime) {
  // new feature
} else {
   // safe fallback
}
```

izaakschroeder:
It would be trivial to write a tiny function for this.

```javascript
fsIs(fs, 'readable', 'xxx')
fsIs(fs, 'mtime');
```

I suppose this function could also do the job of checking member properties/functions (i.e. `readable` implies the existence of `createReadStream`).

dougwilson:
Gotcha. My thought, though, would mean adding the whole `capabilities` information documentation would make the README even longer than ever before, haha. Remember that whatever we do here, everything also has to be fully documented without needing to reference any other document besides the readme to be able to write something you can pass as the `fs` argument, so keep that in mind with the proposals :)

izaakschroeder:
Could we have a separate `doc/fs.md` file maybe? Which would basically be a mirror copy of part of the VFS spec (which is basically a mirror of part of the node spec)?

dougwilson:
> Could we have a separate doc/fs.md file maybe?

Unfortunately probably not. The way we keep modules maintainable is that all documentation is only in the README and all code is only in a single `index.js` file. This allows up to make judgement calls for what should be split off into another module. We don't want to end up with more Express's in these orgs, with enormous surface areas that makes it hard to maintain, haha. So we'd just keep it only in the README.

izaakschroeder:
Alright so going back to defining an interface _other_ than `fs` as one of the options you explored: doing that would mean probably a lot less funny business since now one just has to map `fs` onto this new _simple_ interface. The docs would be smaller and it would be `send` specific.

HOWEVER it introduces the same forward-compatibility problem; whatever interface gets defined, well, that's the interface we're stuck with.

dougwilson:
If we think it would be too big for a README, then it seems like what we would do is not even try to mimic Node.js core `fs` at all and instead split off another module and say that that module's value is what you provide here and nothing else, and we would, as a default, `require()` that module and call it's methods and not `fs` at all. If we go this route, though, it cannot be maintained by a person; we should set up a team and an organization around it (like we have `expressjs`, `pillarjs`, `jshttp`, etc.) so we know we re buying into something that is maintained by a team of people and won't get abandoned.

dougwilson:
Maybe like a `jsfs` organization? It could set out to define a `fs` interface that would be well-defined and sane so people can like build virtual ones (and even maintain an official one for Node.js's `fs` module). People would then build `fsjs` bindings for privileged WebKit, for fake virtual drives, etc. It'd basically be like VFS/FUSE. I'm sure a lot of Node.js modules would buy into that.

izaakschroeder:
So then `send` has two paths? If `jsfs` do this, if `fs` then do that? Since `jsfs` would be a peer? And it would be silly to require people to use some `jsfs` adapter just to use `send`. Ergo, I'm slightly scared about defining an interface that (at least to a large extent) isn't a subset of what node already has.

That being said, some well defined `fs`-like interface seems like a good idea :+1:

dougwilson:
Well, what I'm saying is this module would never `require('fs')` at all--it would `require('jsfs-node')` or similar to use as a default if a user didn't pass in the, I suppose, `vfs` option. This would olve the two main issues: we only have a single code path, since we would only ever use `jsfs` interface and we now have a well-defined this to accept as an option to this library (and as a bonus, we don't have to define that interface here, which is not the case for Node.js `fs` since their docs do not help at all if you want to make a clone--you're going to have to read their source).

dougwilson:
The fact that we cannot even interact with Node.js's `fs` without reading the Node.js source code means that no one has any real hope in being able to create a proper clone of it, and no one would be able to properly construct the value for the `fs` option, as this PR implements it.

izaakschroeder:
Maybe we should take a somewhat similar approach to `readable-stream`? https://github.com/nodejs/readable-stream

dougwilson:
What is the approach that `readable-stream` takes? It's not clear to me what that question means.

dougwilson:
I think that based on where the discussion kind of fizzled out plus the PRs that this module has been getting, I think it's probably still too early in the life of this module to be able to declare a strict set of the `fs` surface area that the module will be using.

If there is ever a group that comes up with the virtual fs concept idealized above, that abstracts these things away, we can always reconsider :)

PR #81, opened 2015-09-19
Check module node.js 4.* to raise in express.js

nooks:


dougwilson:
Please update the AppVeyor configuration as well.

nooks:
AppVeyor does not yet support node.js 4.1

dougwilson:
Lame, lol. They should just pull like Travis CI does :) Using just "4.0" is likely good enough for now, because Node.js with 4 follows semver, so there cannot be any breaking changes between 4.0 and 4.1.

Issue #71, opened 2015-01-20
Transform stream from file before sending


peterbartels:
See my pull request. It would be nice to be able to modify the data from the file stream using transforms. This is helpful when the stored data in the file is a template, encrypted or compressed.

dougwilson:
Hi! I saw it existed yesterday, but a security issue came to light as well, so I just need to get that wrapped up first.

peterbartels:
Hi Dough! Any chance you can have look at my PR?

dougwilson:
This will be accepted, I'm just making the rounds through various projects getting them 0.12 supported; sorry for the delay. In the end, the final version of this will have a slightly different API, but will be able to do exactly what you need and more.

peterbartels:
Cool! I'm looking forward to it.

agubler:
@dougwilson Do you know when this (or equivalent functionality) will be merged?

emilioplatzer:
Hi, I also waiting for this. 

I shall use it for render jade and markdown files. I consider than jade, markdown and other similar file types are also "static" files.

dougwilson:
I haven't forgotten about this at all, everyone, just feature bumps on these Express dependencies are still tied to the Express release cycle, I'm sorry. Slowly but surely this dependence will be lifted, especially after Express 3.x (and thus Connect 2.x) is dropped in July.

alexpineda:
Bump on this. I'd like to transform jsx on request.

PR #82, opened 2015-09-29
Add support for transform option.

TimBarham:
This is based on PR #69 by @peterbartels (see issue #71), but with conflicts resolved (and some typos in README.md fixed).

Adds a `transform` option, which allows the file stream to be transformed before piping it to `res`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "after": "0.8.1",
     "istanbul": "0.3.9",
     "mocha": "2.2.5",
+    "readable-stream": "2.0.2",
     "supertest": "1.0.1"
   },
   "files": [
--- a/README.md
+++ b/README.md
@@ -53,7 +53,9 @@ with a dot, for backward-compatibility.
 
 ##### etag
 
-Enable or disable etag generation, defaults to true.
+Enable or disable etag generation.
+
+Defaults to `true`, unless the `transform` option is set.
 
 ##### extensions
 
@@ -70,8 +72,10 @@ in preferred order.
 
 ##### lastModified
 
-Enable or disable `Last-Modified` header, defaults to true. Uses the file
-system's last modified value.
+Enable or disable `Last-Modified` header. Uses the file system's last modified
+value.
+
+Defaults to `true`, unless the `transform` option is set.
 
 ##### maxAge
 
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -31,6 +31,7 @@
     "after": "0.8.1",
     "istanbul": "0.3.9",
     "mocha": "2.2.5",
+    "replacestream": "~4.0.0",
```
dougwilson: It looks like replacestream is not installable on the default Node.js 0.8 npm. Can you pick a different dependency for the testing, please? Also, please only specify exact versions of dependencies outside of pillarjs/jshttp.

dougwilson:
Awesome!

package.json
TimBarham: Yeah, I'm looking at using an older version of `replacestream` which I think will fix this (a version that has no dependencies specified by caret version).

TimBarham:
Switched to an older version of `replacestream`. Node.js 0.8 is happy now :)

dougwilson:
Gotcha. I would much rather not use some ancient version, because that makes it hard for us to support new versions of Node.js. Can you find a module that is current? Otherwise, I mean, just write a transform stream in the test, because it's pretty simple.

dougwilson:
Basically, just like any module, we always keep our dependencies up-to-date (they are slightly out of date right now, only because we haven't made a commit since those modules released new versions). If we find one of our dependencies no longer work with Node.js 0.8, for example, we replace it with something else that does. If we were to accept the PR as-is it would mean we would right away have a task to figure out what to replace the module with. Ideally, we don't really want to accept PRs that immediately put work on us, I'm sure you'll understand :)

TimBarham:
Heh, no problem - I totally understand. I was just taking the easy way out by using the tests from the PR this was based on :smile:. I have an update that I'll push once it is cleaned up that uses a simple, inline replacement for `replacestream`.

TimBarham:
Updated to use latest readable-stream and a trivial replaceStream implementation.

douglasduteil:
:+1:

albertinad:
This is a cool feature! Is there any plan to integrate it soon to master? Thanks!

TimBarham:
@dougwilson - are there any plans to take this change? I'm happy to do the work to resolve the conflicts that have arisen after 10 months of sitting stagnant, but only if there is a point.

wesleytodd:
I am going to take some liberty here and close this in favor of the much more deeply discussed https://github.com/pillarjs/send/pull/195 so that we can consolidate any conversation going forward. We are not going to try and land this feature soon, as we are focusing on the v5 push for `express` and this is not a breaking change, but I dont want to leave too many open PRs which are unlikely to see progress, especially when they are duplicates in many ways.

Issue #86, opened 2015-12-07
Show symbolic links as redirects


piranna:
Add an option to show the symbolic links as 301 redirections to the file location instead of return the file content.

dougwilson:
We could probably do this. I probably won't get to it, but if you can provide a PR (and please ensure it is cross platform), then I can certainly accept it.

piranna:
I've just done a pull-request to add this feature :-)

PR #87, opened 2015-12-08
Return redirection for symlinks

piranna:
This fixes issue #86

### DIFF
--- a/index.js
+++ b/index.js
@@ -16,7 +16,6 @@ var createError = require('http-errors')
 var debug = require('debug')('send')
 var deprecate = require('depd')('send')
 var destroy = require('destroy')
-var encodeUrl = require('encodeurl')
 var escapeHtml = require('escape-html')
 var etag = require('etag')
 var EventEmitter = require('events').EventEmitter
@@ -905,24 +904,6 @@ SendStream.prototype.setHeader = function setHeader (path, stat) {
 function clearHeaders (res) {
   res._headers = {}
   res._headerNames = {}
-}
-
-/**
- * Collapse all leading slashes into a single slash
- *
- * @param {string} str
- * @private
- */
-function collapseLeadingSlashes (str) {
-  for (var i = 0; i < str.length; i++) {
-    if (str[i] !== '/') {
-      break
-    }
-  }
-
-  return i > 1
-    ? '/' + str.substr(i)
-    : str
 }
 
 /**
--- a/index.js
+++ b/index.js
@@ -80,10 +80,8 @@ var listenerCount = EventEmitter.listenerCount ||
   function (emitter, type) { return emitter.listeners(type).length }
 
 
-function responseStatus(res, code, msg)
-{
-  if(msg == null)
-  {
+function responseStatus(res, code, msg) {
+  if(msg == null) {
     msg = statuses[code]
 
     res.setHeader('Content-Type', 'text/plain; charset=UTF-8')
@@ -95,8 +93,7 @@ function responseStatus(res, code, msg)
   res.end(msg)
 }
 
-function redirect(res, loc)
-{
+function redirect(res, loc) {
   // redirect
   res.setHeader('Content-Type', 'text/html; charset=UTF-8')
   res.setHeader('Location', loc)
--- a/index.js
+++ b/index.js
@@ -490,9 +490,15 @@ SendStream.prototype.redirectSymbolicLink = function redirectSymbolicLink(path) 
   {
     if (err) return self.onStatError(err)
 
+    // Get absolute path on the real filesystem of the destination
+    path = dirname(path)
+    var to = resolve(path, linkString)
+
+    // Check destination is not out of files root
+    if(to.indexOf(self._root) !== 0) return this.error(403)
+
     // Get relative paths for all symlinks, also for absolute ones
-    path       = dirname(path)
-    linkString = relative(path, resolve(path, linkString))
+    linkString = relative(path, to)
 
     // Resolve the URL, and make it relative (is this necessary?)
     linkString = url.resolve(self.path, linkString)
--- a/index.js
+++ b/index.js
@@ -30,7 +30,7 @@ var statuses = require('statuses')
 var Stream = require('stream')
 var util = require('util')
 
-var dirname  = path.dirname
+var dirname = path.dirname
 var relative = path.relative
 var url = require('url')
 
@@ -82,10 +82,9 @@ module.exports.mime = mime
 var listenerCount = EventEmitter.listenerCount ||
   function (emitter, type) { return emitter.listeners(type).length }
 
-
-function responseStatus(res, code, msg) {
+function responseStatus (res, code, msg) {
   var errMsg = msg
-  if(errMsg == null) {
+  if (errMsg == null) {
     errMsg = statuses[code]
 
     res.setHeader('Content-Type', 'text/plain; charset=UTF-8')
@@ -97,7 +96,7 @@ function responseStatus(res, code, msg) {
   res.end(errMsg)
 }
 
-function redirect(res, loc) {
+function redirect (res, loc) {
   // redirect
   res.setHeader('Content-Type', 'text/html; charset=UTF-8')
   res.setHeader('Location', loc)
@@ -107,7 +106,6 @@ function redirect(res, loc) {
 
   responseStatus(res, 301, msg)
 }
-
 
 /**
  * Return a `SendStream` for `req` and `path`.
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -12,31 +12,38 @@
...
+var fresh       = require('fresh')
+var mime        = require('mime')
+var ms          = require('ms');
+var onFinished  = require('on-finished')
+var parseRange  = require('range-parser')
+var statuses    = require('statuses')
+
+var debug     = require('debug')('send')
 var deprecate = require('depd')('send')
-var destroy = require('destroy')
```
dougwilson: Please undo all these changes.

dougwilson:
Hi! Thanks for the pull request! Please undo all the style changes, mainly because they are unwelcome (we weren't asking for style changes) and it makes it really hard for me to understand what changed from looking at the diff.

Also, please add documentation and tests for the new feature so we can accept it :)

index.js
piranna: I just put the requires sorted alphabetically and grouped by build-ins, third-parties and local modules, as it's usually accepted in the Node.js community. Also I've made the requires style uniform, instead of some of them using its own `var` statement and others a common one, and some requires that were hidden between the variables declarations. Do you still want these clean-up changes to be undone?

index.js
```diff
@@ -57,6 +64,35 @@ module.exports.mime = mime
 var listenerCount = EventEmitter.listenerCount
   || function(emitter, type){ return emitter.listeners(type).length; };
 
+
+function responseStatus(res, code, msg)
+{
+  if(msg == null)
```
dougwilson: Please match existing styles, for example the bracket style really stands out as misplaced.

index.js
```diff
@@ -396,28 +427,42 @@ SendStream.prototype.isRangeFresh = function isRangeFresh(){
  * @private
  */
 
-SendStream.prototype.redirect = function redirect(path) {
+SendStream.prototype.redirectDirectory = function redirectDirectory(path) {
```
dougwilson: Changing this public method name is a breaking change, meaning accepting this would require a major version bump. There is no timeline for when a major will happen due to Express support, if that is OK.

index.js
dougwilson: > Do you still want these clean-up changes to be undone?

Yes. It is extremely important to us that git blame stays intact, thus why there is a mixed style.

index.js
```diff
@@ -396,28 +427,42 @@ SendStream.prototype.isRangeFresh = function isRangeFresh(){
...
   if (listenerCount(this, 'directory') !== 0) {
-    this.emit('directory')
-    return
+    return this.emit('directory')
```
dougwilson: Please leave the return on it's own line. We want to return undefined, not the event emitter instance.

dougwilson:
Hi! Your original suggested I agreeded to was "Add an option to show the symbolic links as 301 redirections to the file location instead of return the file content.". I don't see an option to enable the feature. If this always enabled? If not, what is the option? We need an option to enable/disable this feature. If you want it accepted any time soon, you should probably default it to disabled so it's backwards compatible.

dougwilson:
Oh, sorry, looks like you added a test from when I looked at it hours ago, haha. Please try to add more tests, including non happy paths. A new feature should have many tests, not just one, as there are probably edge cases you need to test for, for example, what about folder junctions vs folder symlinks? I wouldn't expect a junction to redirect, is that the case with these changes or no?

piranna:
I've just done your suggestions :-)

index.js
```diff
@@ -396,28 +423,43 @@ SendStream.prototype.isRangeFresh = function isRangeFresh(){
...
+SendStream.prototype.redirect = function redirectDirectory(path) {
   if (listenerCount(this, 'directory') !== 0) {
     this.emit('directory')
     return
   }
 
   if (this.hasTrailingSlash()) {
-    this.error(403)
-    return
+    return this.error(403)
```
dougwilson: This should be returning `undefined`, rather than the event emitter instance.

index.js
```diff
@@ -57,6 +61,32 @@ module.exports.mime = mime
...
+  res.setHeader('X-Content-Type-Options', 'nosniff')
+  res.end(msg)
+}
+
+function redirect(res, loc) {
+  // redirect
+  res.setHeader('Content-Type', 'text/html; charset=UTF-8')
+  res.setHeader('Location', loc)
+
+  loc = escapeHtml(loc)
```
dougwilson: Do not reassign a variable tried into `arguments`, as that results in a v8 deopt. Declare a new variable name, please.

index.js
```diff
@@ -57,6 +61,32 @@ module.exports.mime = mime
 var listenerCount = EventEmitter.listenerCount
   || function(emitter, type){ return emitter.listeners(type).length; };
 
+
+function responseStatus(res, code, msg) {
+  if(msg == null) {
+    msg = statuses[code]
```
dougwilson: Do not reassign a variable tried into `arguments`, as that results in a v8 deopt. Declare a new variable name, please.

index.js
```diff
@@ -57,6 +61,32 @@ module.exports.mime = mime
...
+
+  res.statusCode = code
+  res.setHeader('Content-Length', Buffer.byteLength(msg))
+  res.setHeader('X-Content-Type-Options', 'nosniff')
+  res.end(msg)
+}
+
+function redirect(res, loc) {
+  // redirect
+  res.setHeader('Content-Type', 'text/html; charset=UTF-8')
```
dougwilson: This refactored function is missing the `X-Content-Type-Options` security header the old code was setting. Please add it back :)

index.js
```diff
@@ -396,28 +423,43 @@ SendStream.prototype.isRangeFresh = function isRangeFresh(){
...
+SendStream.prototype.redirectSymbolicLink = function redirectSymbolicLink(path) {
+  var self = this
+
+  fs.readlink(path, function(err, linkString)
+  {
+    if (err) return self.onStatError(err)
+
+    // Get relative paths for all symlinks, also for absolute ones
+    path       = dirname(path)
+    linkString = relative(path, resolve(path, linkString))
```
dougwilson: A symlink can be pointing outside of the `root` set in the constructor. If a redirect occurs like this, the resulting URL is going to be useless is most cases, and in the worse case, the `root` argument is supposed to provide security by boxing all requests into the given root, and the redirect will reveal directory names that could potentially be sensitive. I would expect that the redirect would not work if the symlink pointed outside of the root.

index.js
piranna: I should have missed it before. Anyway it doesn't matter if it returns the event-emitter or undefined as far as it's not being used, and in fact I've seen in several places that this is the recomended way to do it. Anyway, I've already fixed it.

index.js
piranna: Good point. I think here we have three options:

* Leave it as is, and left to the user the decision if the paths show sensitive info
* Return an error (403 forbidden?)
* Return the actual file content

Maybe we could add an option here to check what behaviour to use...

dougwilson: I would say, as a default, a 403 (option 2) may be the best. Option 3 is not great, because it means this feature is no longer "redirect on symlinks" but instead "maybe redirect on symlinks, maybe not; it depends on what your root is, ultimately". Option 1 is probably not a good option to give.

piranna: So seems we have a winner... :-) I proposed the third one just to make it more transparent. What do you think about adding an option between 1 and 2 (being the 403 the default)?

dougwilson: Just option 2 for now.

piranna:
I've just set 403 error by default if destination is outside files root.

piranna:
Any update on this?

dougwilson:
Hi @piranna, sorry, I have not taken a look since you changed the pull request. Can you at least take a look and fix the failing tests so the pull request passes the test while I take a look at the new changes? Also note that one of the tests is marked as "double callback!" with your changes, so even though that's not causing a failure, you'll need to get that fixed as well.

piranna:
Tests are failling to some unrelated test I can't be able to identify what's the source of the problem ("should return 500 on error when reading file"). Could you be able to take a look so this could be merged?

piranna:
The problem that's giving is about the callback being called twice, probably due to the `nextTick()` call in the test.

dougwilson:
There are still a lot of failing tests and you haven't fixed the double callback, either. In addition, it looks like when you merged, you accidentally reverted code, including a security fix. Please fix all that up and ping me and I'll come back and take a look :)!

wesleytodd:
I am going to close this. It seems like a good *idea* but we would need to have a much more clear and clean change set to ever merge this, which seems unlikely at this point. Feel free to reopen a new PR with *just* the minimal changes required to land the feature if you are willing and able to work on this again in the future.

PR #89, opened 2015-12-10
`index` event

piranna:
Allow to render a custom `index` page from a function.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,9 +81,14 @@ This is skipped if the requested file already has an extension.
 
 ##### index
 
-By default send supports "index.html" files, to disable this
-set `false` or to supply a new index pass a string or an array
-in preferred order.
+By default *send* supports "index.html" files. To disable this, set index option
+to `false`, or you can be able to supply a new index passing a string with the
+filename or an array in the preferred order.
+
+Alternatively, you can pass a function that will generate the index. The
+function will receive the filesystem absolute path and will have the
+`SendStream` instance as `this` object. You should call internally to the
+`this.send()` method or try to mimic its behaviour.
 
 ##### lastModified
 
@@ -229,7 +234,7 @@ var app = http.createServer(function onRequest (req, res) {
 }).listen(3000)
 ```
 
-## License 
+## License
 
 [MIT](LICENSE)
 
--- a/index.js
+++ b/index.js
@@ -144,11 +144,9 @@ function SendStream (req, path, options) {
 
   this._extensions = normalizeList(opts.extensions, 'extensions option')
 
-  this._index = opts.index instanceof Function
-    ? opts.index
-    : opts.index !== undefined
-      ? normalizeList(opts.index, 'index option')
-      : ['index.html']
+  this._index = opts.index !== undefined
+    ? normalizeList(opts.index, 'index option')
+    : ['index.html']
 
   this._lastModified = opts.lastModified !== undefined
     ? Boolean(opts.lastModified)
@@ -707,8 +705,20 @@ SendStream.prototype.sendFile = function sendFile (path) {
  * @api private
  */
 SendStream.prototype.sendIndex = function sendIndex (path) {
-  var i = -1
-  var self = this
+  if (listenerCount(this, 'index')) {
+    this.emit('index', path)
+    return
+  }
+
+  // We don't want to render an index, process it as a `file`. Probably it will
+  // return a 404 error.
+  if (!this._index.length) {
+    this.sendFile(path);
+    return this.res;
+  }
+
+  var i = -1;
+  var self = this;
 
   function next (err) {
     if (++i >= self._index.length) {
--- a/index.js
+++ b/index.js
@@ -541,8 +541,8 @@ SendStream.prototype.pipe = function pipe (res) {
 
   // index file support
   if (this._index instanceof Function) {
-    this._index(path);
-    return res;
+    this._index(path)
+    return res
   }
   if (this._index.length && this.path[this.path.length - 1] === '/') {
     this.sendIndex(path)
@@ -713,12 +713,12 @@ SendStream.prototype.sendIndex = function sendIndex (path) {
   // We don't want to render an index, process it as a `file`. Probably it will
   // return a 404 error.
   if (!this._index.length) {
-    this.sendFile(path);
-    return this.res;
+    this.sendFile(path)
+    return this.res
   }
 
-  var i = -1;
-  var self = this;
+  var i = -1
+  var self = this
 
   function next (err) {
     if (++i >= self._index.length) {
--- a/index.js
+++ b/index.js
@@ -145,7 +145,9 @@ function SendStream (req, path, options) {
   this._extensions = normalizeList(opts.extensions, 'extensions option')
 
   this._index = opts.index !== undefined
-    ? normalizeList(opts.index, 'index option')
+    ? (opts.index instanceof Function
+      ? opts.index
+      : normalizeList(opts.index, 'index option'))
     : ['index.html']
 
   this._lastModified = opts.lastModified !== undefined
--- a/index.js
+++ b/index.js
@@ -144,11 +144,18 @@ function SendStream (req, path, options) {
 
   this._extensions = normalizeList(opts.extensions, 'extensions option')
 
-  this._index = opts.index !== undefined
-    ? (opts.index instanceof Function
-      ? opts.index
-      : normalizeList(opts.index, 'index option'))
-    : ['index.html']
+  switch (typeof opts.index) {
+    case 'function':
+      this._index = opts.index
+      break
+
+    case 'undefined':
+      this._index = ['index.html']
+      break
+
+    default:
+      this._index = normalizeList(opts.index, 'index option')
+  }
 
   this._lastModified = opts.lastModified !== undefined
     ? Boolean(opts.lastModified)
@@ -542,7 +549,7 @@ SendStream.prototype.pipe = function pipe (res) {
   }
 
   // index file support
-  if (this._index instanceof Function) {
+  if (typeof this._index === 'function') {
     this._index(path)
     return res
   }
--- a/index.js
+++ b/index.js
@@ -549,6 +549,7 @@ SendStream.prototype.pipe = function pipe (res) {
     return res
   }
 
+  // Render a file, or an index as a file
   this.sendFile(path)
   return res
 }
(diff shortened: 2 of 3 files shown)


piranna:
Any update on this?

index.js
```diff
@@ -397,8 +395,8 @@ SendStream.prototype.isRangeFresh = function isRangeFresh(){
  */
 
 SendStream.prototype.redirect = function redirect(path) {
-  if (listenerCount(this, 'directory') !== 0) {
-    this.emit('directory')
+  if (listenerCount(this, 'directory')) {
+    this.emit('directory', path)
```
dougwilson: Adding a new argument to the `directory` event does not seem related to adding an `index` event. Can you please restructure this pull request to only do what was described in the pul request summary and ping me when you have updated it?

piranna:
I've just rebased my changes over master and tests are passing, would you be able to merge it?

dougwilson:
The comment I made on your code hasn't been addressed yet. Please address it and I'll take another look :)!

piranna:
> The comment I made on your code hasn't been addressed yet. Please address it and I'll take another look :)!

Sorry, I missed the comment :-P I've just fixed it and move the `path` argument change on a [different pull-request](https://github.com/pillarjs/send/pull/115) :-)

piranna:
Code rebased and tests passing. Can we focus on that over the other pull-requests? This is the most important for NodeOS at this moment...

dougwilson:
Sure! Can you use typeof instead of instanceof for function checks? Why do we need both an option and an event? Is it not possible to just have one or the other?

piranna:
At first I added an event like 'directory' and so, but later found it was
easier to use an optional callback. In fact, if I would redesign the API, I
would remove all the events except error and just use optional callbacks,
it doesn't makes sense to be able to register several functions that can
return a value for a single input (the path)...

piranna:
I have just changed `instanceof` for `typeof`.

dougwilson:
Cool! Why do we need both an option and an event? Is it not possible to just have one or the other? If we can use events over function options that would better fit in the current API. I know the API may not be what everyone is used to, but it is what I have inherited, and it doesn't sound like you want to wait for a discussion on changing the way the API works to land this feature :) At the very least, you need to move the test you made for the index event out of the section for the index option and add documentation for that to the README as well.

index.js
```diff
@@ -142,13 +142,20 @@ function SendStream (req, path, options) {
     this._dotfiles = undefined
   }
 
-  this._extensions = opts.extensions !== undefined
-    ? normalizeList(opts.extensions, 'extensions option')
```
dougwilson: Is this a necessary change for this PR?

README.md
```diff
@@ -229,7 +234,7 @@ var app = http.createServer(function onRequest (req, res) {
 }).listen(3000)
 ```
 
-## License
```
dougwilson: Is this a necessary change for this PR?

test/send.js
```diff
@@ -702,9 +702,11 @@ describe('send(file).pipe(res)', function () {
         .pipe(res)
       })
 
+      var filePath = path.join(fixtures, 'pets', 'index.html')
+
       request(app)
       .get('/pets/')
-      .expect(200, fs.readFileSync(path.join(fixtures, 'pets', 'index.html'), 'utf8'), done)
+      .expect(200, fs.readFileSync(filePath, 'utf8'), done)
```
dougwilson: Is this a necessary change for this PR?

index.js
```diff
@@ -542,11 +549,16 @@ SendStream.prototype.pipe = function pipe (res) {
   }
 
   // index file support
+  if (typeof this._index === 'function') {
```
dougwilson: Can this check just be moved into the sendIndex method? It is a public method, and it's weird the event emit is in there but not this check.

README.md
```diff
@@ -81,9 +81,14 @@ This is skipped if the requested file already has an extension.
 
 ##### index
 
-By default send supports "index.html" files, to disable this
-set `false` or to supply a new index pass a string or an array
-in preferred order.
+By default *send* supports "index.html" files. To disable this, set index option
```
dougwilson: Since this section is about the index option, I don't think you need to refer to itself, as it ends up being really wordy.

README.md
```diff
@@ -81,9 +81,14 @@ This is skipped if the requested file already has an extension.
...
+to `false`, or you can be able to supply a new index passing a string with the
```
dougwilson: "you can be able to" doesn't sound right. I think probably just remove those words?

dougwilson: And passing would become just pass

README.md
```diff
@@ -81,9 +81,14 @@ This is skipped if the requested file already has an extension.
...
+filename or an array in the preferred order.
+
+Alternatively, you can pass a function that will generate the index. The
```
dougwilson: Rather than tacking this on as a separate paragraph, can you incorporate them all together? Otherwise, it sounds very hack in, but having a complete paragraph saying that the option only takes a boolean, string, or array, then another paragraph following to say "but wait! We forgot to add it takes a function too"

README.md
```diff
@@ -81,9 +81,14 @@ This is skipped if the requested file already has an extension.
...
+function will receive the filesystem absolute path and will have the
+`SendStream` instance as `this` object. You should call internally to the
```
dougwilson: Having the "this" populated on a callback like function is very odd, and very easy to get lost. It is kind of annoying already for events, but typically I never see this on straight functions. Can everything a user would need just be passed as arguments?

README.md
```diff
@@ -81,9 +81,14 @@ This is skipped if the requested file already has an extension.
...
+`this.send()` method or try to mimic its behaviour.
```
dougwilson: Can the docs be clearer on how someone is supposed to "mimic its behavior"?

README.md
piranna: That's formal english... As you want.

dougwilson:
And I was jus able to sit down a s review your pull request, woot! We can get this moving along for sure :) I just had a few questions and comments; some in this conversation itself and some inline in the code.

README.md
piranna: This clean-up is done automatically by Atom. Correct markdown don't accept blanks at the end of the line, except two blank spaces to force to do backline.

index.js
piranna: `normalizeList()` now return an empty list when the input is `undefined`, so the check here is bloated.

index.js
piranna: I would in fact remove the `index` event at all... This would be moved there too, but then the check for the function would need to be added both on the next `if` conditional AND inside the `sendIndex()` function too.

test/send.js
piranna: ESlint was giving me errors of lines too long, this fixed it. Source code lines should never be longer than 80 columns.

index.js
dougwilson: I don't see any changes to  normalizeList  in your PR. Perhaps I should be clearer: don't include changes in your PR that are not accomplishing your goal. These types of changes need to be different PRs. Please revert this change.

test/send.js
dougwilson: Revert the change if you would like this landed... Please don't include unrelated changes. If we wanted to enforce a max line length, we would do so, but we do not.

index.js
dougwilson: I'm sorry, I don't understand your response. Can you say it in a different way?

README.md
dougwilson: > That's formal english... As you want.

I'm sorry, I can't seem to figure out what your response means...

piranna:
> Cool! Why do we need both an option and an event? Is it not possible to just have one or the other?

Yes, we could use just one of them, and in fact it would be better to use optional callbacks only, I just added the event to be on pair with the API, but I would remove them.


> If we can use events over function options that would better fit in the current API. I know the API may not be what everyone is used to, but it is what I have inherited, and it doesn't sound like you want to wait for a discussion on changing the way the API works to land this feature :)

I'm open for a discussion about changing how the API works, it's just that this seemed to be ready to merge and wanted to use upstream code ASAP. I think events-based API should be deprecated and use optional callbacks instead, it's the correct behaviour and what everybody expects:

* just one and only one listener: callback
* zero or more listeners
  * just one and only one call: promise
  * zero or more calls: event

This also makes the API more explicit about what can and what can't do? What do you think?


> At the very least, you need to move the test you made for the index event out of the section for the index option and add documentation for that to the README as well.

Where should I put the test for the `index` event in case we still use it?

dougwilson:
We can discuss a new API style for this module, absolutely, but no new API style will be merged for months to come, because this module is locked to the release schedule of Express.js itself. Even this new feature will not get released until a minor release of Express.Js is scheduled. I wish we had all this conversation 3 months ago,when I started commenting on the changes, as it would have ended up with the 4.14 Express.js release.

I don't have time this weekend to discuss and think about an entire change to the API of this module, only have been able to sneak in reviewing these changes as a simple feature addition, as it sounded like you were interested in landing. I'll try to circle back to this PR another day when I have some time to discuss a new API :)

piranna:
> this module is locked to the release schedule of Express.js itself

That's totally insane, being so much specific it should have its own lifecycle, specially for this backward compatible things... And if not, semantic versioning is just for that...


> I'll try to circle back to this PR another day when I have some time to discuss a new API :)

Ok, send me an email when you have time to discuss it.

dougwilson:
> That's totally insane, being so much specific it should have its own lifecycle

You're always welcome to use another module or fork this module. This is not up for discussion, I'm sorry. This code could have easily just been embedded into Express core. It is broken out for you to use it without Express, not to have a different release schedule, I'm sorry this is an inconvenience to you. We barely have enough people to maintain all this module, let alone for them all yo have different release timelines,meaning we have to create multiple lineages for modules when security vulnerabilities come out.

I feel like this is getting a little too headed foe me, so I'm not longer going to participate in this PR for time time being. Another member of the organization is welcome to pick up moving this PR forward...

piranna:
I'm currently using a fork of this repo, the point of willing my changes to be merged is to clean up and have a consistent repo, but just by suggesting to use a repo, although acceptable, only contribute to degrade npm ecosystem. On NodeOS we lack a lot of people too, being so picky with the pull-requests just makes everybody losing their time and willing to not contribute anymore. Everybody lose :-(

dougwilson:
Your PR is just never going to land, as you have not even hinted that you are willing to make any changes to your PR, and our philosophies for maintain seems to be too different: You suggest that I just merge anything without question, and I only want to merge quality code. These are unfortunately mutually exclusive. If you want to get a feature added, please work to add the feature with the API we have, rather than debate which API style is better.

piranna:
I did all the changes that you tell me to do. You are not looking for quality code, but a clean and verbose commits history. If you really wanted to have quality code, you'll start by thinking about changing the API, but I only hinted about that and tried to adapt to the current code style and project organization. If you are not willing to maintain this project is ok, just give me permissions and I'll maintain it in your place.

dougwilson:
> I did all the changes that you tell me to do.

I'm not sure why you say this, as there are many, many open comments on your code above, and no commits since any of them. No changes I can see.

> just give me permissions and I'll maintain it in your place

No.

Issue #91, opened 2015-12-12
actual treatment of dotfiles and dotdirs


coolaj86:
This is the actual treatment of docfiles, which differs slightly from the documentation

**default**:

* **200**: http://example.com/.hidden/foo
* **404**: http://example.com/.hidden/.dot
* **200**: http://example.com/.hidden/.deep/foo
* **404**: http://example.com/.hidden/.deep/.dot

`ignore`:

* **404**: http://example.com/.hidden/foo
* **404**: http://example.com/.hidden/.dot
* **404**: http://example.com/.hidden/.deep/foo
* **404**: http://example.com/.hidden/.deep/.dot

`allow`:

* **200**: http://example.com/.hidden/foo
* **200**: http://example.com/.hidden/.dot
* **200**: http://example.com/.hidden/.deep/foo
* **200**: http://example.com/.hidden/.deep/.dot

`deny`:

* **403**: http://example.com/.hidden/foo
* **404**: http://example.com/.hidden/.dot
* **403**: http://example.com/.hidden/.deep/foo
* **404**: http://example.com/.hidden/.deep/.dot

dougwilson:
> which differs slightly from the documentation

Can you be more specific?

coolaj86:
Intuitively (and others may feel differently) I would expect this behavior:

* `undefined`: console.warn('[deprecated] please specify opts.dotfiles') and support legacy to not break code
* `true` or `'allow'`: no special treatment
* `false` or `'deny'`: 403 all dotfiles and 403 dotdirs
* `['.allow-me', '.me-too']`: allow these specific dotfiles / dotfolders
* ``[]` or 'ignore'`: 404 all dotfiles and 404 dotdirs

dougwilson:
>  undefined: console.warn('[deprecated] please specify opts.dotfiles') and support legacy to not break code

It's not deprecated, thus why there is no warning. Now sure why we'd want to deprecate this.

dougwilson:
For your `deny` example, I get 403 for all 4 cases you provided above. Can you please share your code so I can reproduce? They should all be 403 and they all are for me.

coolaj86:
The code says it's `legacy`: https://github.com/pillarjs/send/blob/master/index.js#L108

When I hear `legacy` I think "we wish we had done it differently, but it is how it is" which I further translate to: "let's deprecate and remove this in the next major version" - which is what express itself has done several times.

[...]

PR #92, opened 2015-12-12
address #91: document actual behavior

coolaj86:


README.md
```diff
@@ -39,13 +39,14 @@ not the actual file-system path).
...
 the path itself without checking if the path actually exists on the
-disk. If `root` is specified, only the dotfiles above the root are
+disk. If `root` is specified, only dotfiles below the root are
 checked (i.e. the root itself can be within a dotfile when when set
 to "deny").
 
-  - `'allow'` No special treatment for dotfiles.
-  - `'deny'` Send a 403 for any request for a dotfile.
-  - `'ignore'` Pretend like the dotfile does not exist and 404.
+  - `undefined` dotdirs have no special treadment, dotfiles are 404'd.
```
dougwilson: Typo in "treadment"

dougwilson:
The term "dotdirs" is not defined in the README, and in the README, "dotfiles" is defined to mean both files and directories. What is a "dotdirs"? Please update the definition of "dotfiles" and/or define what a "dotdirs" is and how they differ, exactly.

dougwilson:
Also, for the dotfiles/dotdirs, keep in mind as the documentation says, we literally do not stat the disk. This means that for the path `/.hidden/.dot` this module does not have any idea if `.dot` is a file or a directory and will treat it exactly the same as far as `dotfiles` options goes. This is also why we only have one option `dotfiles` and not split between  a `dotfiles` and a `dotdirs` option.

dougwilson:
@coolaj86 , please ping me here once you address the issues and update the PR :)

coolaj86:
Yes, that makes sense... I'll double check my routing foo and see if I can figure out why it appears that dotfiles nested under dotdirs behave differently than leaf node's.

I'll go ahead and re-PR with your suggestions.

README.md
dougwilson: Please, absolutely remove `undefined` as an option; this is confusing and will confuse a lot of people. People do not provide an option with the value of `undefined` under normalize circumstances; the code in the module looks for that value, yes, because it's what it means to not have supplied a value in JavaScript.

dougwilson:
> I'll go ahead and re-PR with your suggestions.

Please just update this PR, don't make a new one, please...

coolaj86:
created https://github.com/pillarjs/send/pull/94 before I saw this

PR #94, opened 2015-12-12
Clarify dotfile documentation

coolaj86:
##### dotfiles

Set how "dotfiles" are treated when encountered. A dotfile is a file
or directory that begins with a dot (".").

Note: This check is only performed on the path (of the url).
It doesn't affect `root` (of the filesystem) and doesn't actually check for dotfiles on disk.

  - **default behavior** Ignore child dotfiles, but treat parents normally.
    - 404 `/.well-known`, `/.foo/.bar`
    - 200 `/.well-known/existing-file`
  - `'allow'` No special treatment for dotfiles.
    - 200 `/.exists`
    - 404 `/.doesnt-exist`
  - `'deny'` Send a 403 for any request for a dotfile.
    - 403 `/.exists`, `/.doesnt-exist`
  - `'ignore'` Pretend like the dotfile does not exist and 404.
    - 404 `/.exists`, `/.doesnt-exist`

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,19 +37,21 @@ not the actual file-system path).
 ##### dotfiles
 
 Set how "dotfiles" are treated when encountered. A dotfile is a file
-or directory that begins with a dot ("."). Note this check is done on
-the path itself without checking if the path actually exists on the
-disk. If `root` is specified, only the dotfiles above the root are
-checked (i.e. the root itself can be within a dotfile when when set
-to "deny").
+or directory that begins with a dot (".").
 
+Note: This check is performed on the url pathname.
+It doesn't affect `root` and doesn't actually check if the path existence on disk.
+
+  - `undefined` (legacy) Ignore child dotfiles, but treat parents normally
+    - 404 `/.well-known`, `/.foo/.bar`
+    - 200 `/.well-known/existing-file`
   - `'allow'` No special treatment for dotfiles.
+    - 200 `/.exists`
+    - 404 `/.doesnt-exist`
   - `'deny'` Send a 403 for any request for a dotfile.
+    - 403 `/.exists`, `/.doesnt-exist`
   - `'ignore'` Pretend like the dotfile does not exist and 404.
-
-The default value is _similar_ to `'ignore'`, with the exception that
-this default will not ignore the files within a directory that begins
-with a dot, for backward-compatibility.
+    - 404 `/.exists`, `/.doesnt-exist`
 
 ##### etag
 
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ or directory that begins with a dot (".").
 Note: This check is performed on the url pathname.
 It doesn't affect `root` and doesn't actually check if the path existence on disk.
 
-  - `undefined` (legacy) Ignore child dotfiles, but treat parents normally
+  - **default behavior** Ignore child dotfiles, but treat parents normally.
     - 404 `/.well-known`, `/.foo/.bar`
     - 200 `/.well-known/existing-file`
   - `'allow'` No special treatment for dotfiles.
--- a/README.md
+++ b/README.md
@@ -39,8 +39,8 @@ not the actual file-system path).
 Set how "dotfiles" are treated when encountered. A dotfile is a file
 or directory that begins with a dot (".").
 
-Note: This check is performed on the url pathname.
-It doesn't affect `root` and doesn't actually check if the path existence on disk.
+Note: This check is only performed on the path (of the url).
+It doesn't affect `root` (of the filesystem) and doesn't actually check for dotfiles on disk.
 
   - **default behavior** Ignore child dotfiles, but treat parents normally.
     - 404 `/.well-known`, `/.foo/.bar`


dougwilson:
Please incorporate all my feedback from the former PR. I asked not to list `undefined` as an actual option, or I wouldn't accept the changes. A lot of people use JSON to store configuration and they will continually ask how to specify "undefined" in JSON. I had a module like that and the issues were endless around that point. Please ping me once you removed that from this PR so I can take another look :)

README.md
```diff
@@ -37,19 +37,21 @@ not the actual file-system path).
...
 
 Set how "dotfiles" are treated when encountered. A dotfile is a file
-or directory that begins with a dot ("."). Note this check is done on
-the path itself without checking if the path actually exists on the
-disk. If `root` is specified, only the dotfiles above the root are
-checked (i.e. the root itself can be within a dotfile when when set
-to "deny").
+or directory that begins with a dot (".").
 
+Note: This check is performed on the url pathname.
```
dougwilson: Can you define what "url pathname" means here? Currently there is nothing in the documentation that makes reference to what the url pathname would be. Before It just said "path", which was the argument name to the constructor, which was at least continuous. Would it makes sense to keep saying "path" or just define what "url pathname" is and how it relates to the arguments the user specified?

coolaj86:
sorry, lot's of text flying back and forth, missed that note.

Changed it to **default behavior**. Perhaps allowing `null` would be a good option? (looking at the code it seems that `if (!opt.dotfiles) {}` should work just fine.

I find it very intuitive when libraries treat `undefined`, *not defined*, and `null` all the same.

README.md
```diff
@@ -37,19 +37,21 @@ not the actual file-system path).
...
+It doesn't affect `root` and doesn't actually check if the path existence on disk.
```
dougwilson: Can you clarity what "It doesn't affect `root`" means? To me, as even the author, this statement doesn't make sense. If it did "affect root", I would presume that it would have altered that property's value?

coolaj86:
updated

just trying to avoid the confusion of the ambiguous term `path` which refers only to the url in the case of this option, but also could refer to disk, generally

I think I found a better way to state it which is even more clear

dougwilson:
> Perhaps allowing null would be a good option? (looking at the code it seems that if (!opt.dotfiles) {} should work just fine.

No, because we are not altering the behavior based on what the value _is_, but rather the _absence_ of the value_. JavaScript has defined `undefined` to equal not defined, and not `null`. You can see this reenforced very hard in ES6 with default argument values; passing `undefined` will end up with the default value but `null` will be `null`. Eventually when we use ES6 and want to use the built-in mechanisms for defaults, we don't want it to require an API change.

README.md
```diff
@@ -37,19 +37,21 @@ not the actual file-system path).
...
 Set how "dotfiles" are treated when encountered. A dotfile is a file
-or directory that begins with a dot ("."). Note this check is done on
-the path itself without checking if the path actually exists on the
-disk. If `root` is specified, only the dotfiles above the root are
-checked (i.e. the root itself can be within a dotfile when when set
-to "deny").
+or directory that begins with a dot (".").
 
+Note: This check is only performed on the path (of the url).
+It doesn't affect `root` (of the filesystem) and doesn't actually check for dotfiles on disk.
```
dougwilson: Now "It doesn't affect `root` (of the filesystem)" literally says to me "It doesn't affect / (or C:\ on Windows), because that is the root of the file system. Seeing a line like that would kind of make me worry (wait, if it doesn't affect the root of my filesystem, what would?). A callout like that would imply that there is something that does the opposite somewhere.

coolaj86:
So how would one explicitly declare "I would like the default value, please"?

I'm a big fan of the python zen: explicit over implicit so in my code and config I tend to declare the default values (sometimes with the comment `//default`).

This helps me in that A) I know the option is something that can be changed and B) I know I intended that behavior rather than just being neglectful.

README.md
coolaj86: Yeah, I can see what you mean.

I'll let take it from here.

dougwilson:
> So how would one explicitly declare "I would like the default value, please"?

In JavaScript, you don't provide the value. For example, do not include the key for it in your object. There is a lot of code that does the defaults using the `in` operator, but that is slightly misguided since the JS spec says to strict equal to `undefined` to use the default. Typically, the target audience for a JS module is users that understand JS. I like to just follow the specs set out in JavaScript itself (ECMA-262). Saying "the default value" should get people to say "ah, it's just like how defaults work in the JavaScript language" so I don't have to go describing how our defaults mechanism is different from the language on every module.

README.md
dougwilson: Cool. Once we get this into a position we feel good about, since this is a later large change with formatting as well, I usually ask one of my documentation writer co-workers to look it over. I may have more feedback in the coming weeks based on her feedback :)

dougwilson:
I'm closing this pull request, as it may have gone stale, as there are actions waiting to be done on the code for over a month now. If you are interested in picking it back up, let me know and I can always re-open the pull request!

PR #97, opened 2016-01-20
Add defaultType option (default NONE)

felicienfrancois:
The default value for the `content-type` header when send failed to
resolve it from file path.
By default, the `content-type` header is not set.


index.js
```diff
@@ -137,6 +137,8 @@ function SendStream(req, path, options) {
   if (!this._root && opts.from) {
     this.from(opts.from)
   }
+
+  mime.default_type = opts.defaultType || null;
```
dougwilson: The `mime` variable is a static variable, shared between all instances of `send`. Setting this off an option does not make sense, because executing two `send` commands with two different `defaultType` options will result in both requests having the `defaultType` that was send in the second call to `send`.

felicienfrancois: You're right, thanks for pointing me out this issue, I'll update this

Does it means that currently send may be affected when mime.default_type is set in another module (and possibly breaks if it is null)?

felicienfrancois:
I updated this pull request to the master branch with:
1°) Default content-type is now `text/plain` if defaultType is not set
2°) Uses a dedicated mime instance which share mime map data with the shared one but not the default_type
3°) Added tests

@dougwilson
I'm not sure about the expected behavior for the mime instance.
I "cloned" the mime module into a _mime instance which share mime map with the "global" one but not share the default_type (which is set to null). This instance is still declared as static and so it should be shared by all instance of send. It is also the one which is exported on send.mime (should it be ?).
Then the defaultType option is stored in each instance of send and is used when the result of _mime.lookup is null.

Issue #80, opened 2015-09-17
A named anonymous function issue on Safari's strict mode


ajhsu:
In strict mode, Safari 8.0.8 on OS X 10.10.5 raises an exception if a named function's name is the same as one of the function's argument names.

For example: at index.js:L242
`SendStream.prototype.error = function error(status, error) {`

will cause an error on Safari ( but not on Chrome and Firefox ):
`SyntaxError: Cannot declare a parameter named 'error' as it shadows the name of a strict mode function.`

Any sugguestion?

dougwilson:
Hi! Yes, that is definitely an issue, as this module does not have any support for web browsers, and would never be expected to function in Safari. The problem is that if there were a true strict mode violation, then we would alter it, but it's actually a known bug in Safari's implementation of strict mode. The problem is that if we cannot add a test that fails with this and passes after the change, then we cannot support altering it, as it'll just get altered back again at some point by another maintainer not knowing.

Now, perhaps the suggested course of action is to get this module to function in web browsers like Safari, contribute CI setup to test this module in Safari and other browsers, and bring web browsers to be supported by this module, if you are up to the task :)

dougwilson:
Hi @ajhsu , I assume by the last of response over the months you're not up to the task currently. That's OK! I'm going to close the issue, as we just don't have the resources to support web browsers at the moment, but if you ever get the time to donate your time & effort to add the support, just let us know!

PR #99, opened 2016-01-30
Rename parameter to fix strict mode violation

koluch:
When this module loaded into Safari (being included to bundle as a dependency), it cause following error:

"Cannot declare a parameter named 'error' as it shadows the name of a strict mode function."

So, to fix it I renamed 'error' parameter in 'error' function

### DIFF
--- a/index.js
+++ b/index.js
@@ -239,10 +239,10 @@ SendStream.prototype.maxage = deprecate.function(function maxage(maxAge) {
  * @private
  */
 
-SendStream.prototype.error = function error(status, error) {
+SendStream.prototype.error = function error(status, err) {
   // emit if listeners instead of responding
   if (listenerCount(this, 'error') !== 0) {
-    return this.emit('error', createError(error, status, {
+    return this.emit('error', createError(err, status, {
       expose: false
     }))
   }


dougwilson:
Hi @koluch, that being a strict mode violation is actually a bug (since fixed) in Safari's strict mode. Typically this module is designed only to run on the server-side, like in Node.js. Can you ellaborate on how exactly you re using this module inside Safari?

dougwilson:
You can also find a previous conversation about this in https://github.com/pillarjs/send/issues/80

The problem is that if we cannot add a test that fails with this and passes after the change, then we cannot support altering it, as it'll just get altered back again at some point by another maintainer not knowing.

Now, perhaps the suggested course of action is to get this module to function in web browsers like Safari, contribute CI setup to test this module in Safari and other browsers, and bring web browsers to be supported by this module, if you are up to the task :)

koluch:
Hi @dougwilson,

I don't actually remember why did I need this module in my browser - it was a dependency to some another module which I don't use anymore. So, since such fixes are against the workflow of your repository, I will just close this issue.

Anyway, thanks for your response :)

dougwilson:
Ah, I see. I'm glad your issue has been resolved, though! If you ever find yourself in a similar situation, where a module is being bundled in your browser build but you don't actually need it, I'm sure your bundler can be set to exclude certain modules.

PR #100, opened 2016-02-01
Support custom fs (tailing)

ronag:
This allows one to provide a custom fs implementation to send with .e.g. tailing (growing) support.

### DIFF
--- a/index.js
+++ b/index.js
@@ -83,6 +83,7 @@ function send(req, path, options) {
 function SendStream(req, path, options) {
   var opts = options || {}
 
+  this.fs = opts.fs || fs
   this.options = opts
   this.path = path
   this.req = req
@@ -559,6 +560,9 @@ SendStream.prototype.send = function(path, stat){
 
     // unsatisfiable
     if (-1 == ranges) {
+      if (stat.growing) {
+        return this.error(404);
+      }
       debug('range unsatisfiable');
       res.setHeader('Content-Range', 'bytes */' + stat.size);
       return this.error(416);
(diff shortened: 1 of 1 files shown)


dougwilson:
Hi @ronag, thanks for your pull request! If you can, please check out the tests, as there are a bunch of failures with these changes.

I also don't see anything about a `.growing` property in the Node.js documentation: https://nodejs.org/api/fs.html#fs_class_fs_stats what does it mean and where does it come from?

In addition, I'm not sure about this module ever sending back a 202, because it is described as the following in RFC 7231:

> The 202 (Accepted) status code indicates that the request has been accepted for processing, but the processing has not been completed. The request might or might not eventually be acted upon, as it might be disallowed when processing actually takes place.  There is no facility in HTTP for re-sending a status code from an asynchronous operation.
>
> The 202 response is intentionally noncommittal.  Its purpose is to allow a server to accept a request for some other process (perhaps a batch-oriented process that is only run once per day) without requiring that the user agent's connection to the server persist until the process is completed.  The representation sent with this response ought to describe the request's current status and point to (or embed) a status monitor that can provide the user with an estimate of when the request will be fulfilled.

To me, that describing the accepting of the given request (i.e. the `GET` or `HEAD` request that this module will respond to), rather than indicate that some process is still pending (i.e. that you're checking to see if something is done yet).

dougwilson:
Hi @ronag, I'm closing this due to no response in over 2 months. If you get back to me on my questions/comments above, I'll happily re-open this pull request!

PR #104, opened 2016-03-03
Update README.md

### LINKED ISSUE #103: Documented start/end options
I wanted to serve only parts of a large file and found the undocumented start/end options to serve exactly this purpose. E.g. I have a large audio file and only want to stream a short extract of which I know the start and end bytes. So I call `/large.mp3?start=1234&end=5678` and pass the query parameters as options to send and I get exactly what I want.
The start and end options are not documented though, therefore I would like to know if this is just missing documentation, or if they are internal properties and setting them might break other features, e.g. conditional-GET etc.

mh-cbon:
fix #103

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,10 @@ The default value is _similar_ to `'ignore'`, with the exception that
 this default will not ignore the files within a directory that begins
 with a dot, for backward-compatibility.
 
+##### end
+
+Byte offset at which the stream ends.
+
 ##### etag
 
 Enable or disable etag generation, defaults to true.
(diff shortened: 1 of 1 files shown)


tunniclm:
Nice

mh-cbon:
much better indeed

maenu:
Sweet! Thanks guys.

dougwilson:
No problem, @maenu! @mh-cbon jumped right in there and helped you out :) I hope having it documented makes you feel more comfortable about using it :)

maenu:
It absolutely makes me feel more comfortable to have documentation about the features I am using. This makes sure that I am not programming against an implementation detail but a published contract one can put the finger on, should the documentation not be reflected in the runtime behaviour.

mh-cbon:
just needed to be done for everyone : s cool it s done :  )

PR #105, opened 2016-03-19
added accepted options for etag

retep007:
This PR adds funcionality for setting etags, 'weak' 'strong' custom function, without breaking back compatibility.

Needed for https://github.com/expressjs/express/pull/2936

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,9 @@ the 4th byte in the stream.
 
 ##### etag
 
-Enable or disable etag generation, defaults to true.
+Accepts `'weak'`, `'strong'`, custom function `function(data){ return etag}`,
+`true` for weak, `false` for disabled.
+Default is true.  
 
 ##### extensions
 
--- a/index.js
+++ b/index.js
@@ -88,7 +88,7 @@ function SendStream(req, path, options) {
   this.req = req
 
   this._etag = opts.etag !== undefined
-    ? Boolean(opts.etag)
+    ? opts.etag
     : true
 
   this._dotfiles = opts.dotfiles !== undefined
@@ -768,7 +768,12 @@ SendStream.prototype.setHeader = function setHeader(path, stat){
   }
 
   if (this._etag && !res.getHeader('ETag')) {
-    var val = etag(stat)
+    var val = null;
+    if(typeof this._etag == 'function') {
+        val = this._etag(stat);
+    } else {
+        val = etag(stat, {weak: this._etag == 'weak' || this._etag === true})
+    }
     debug('etag %s', val)
     res.setHeader('ETag', val)
   }
--- a/test/send.js
+++ b/test/send.js
@@ -102,6 +102,84 @@ describe('send(file).pipe(res)', function(){
     .expect('etag', /^W\/"[^"]+"$/)
     .end(done);
   })
+  
+  it('should add an ETag strong header field', function(done){
+    var app = http.createServer(function(req, res){
+    function error(err) {
+        res.statusCode = err.status;
+        res.end(http.STATUS_CODES[err.status]);
+    }
+
+    function redirect() {
+        res.statusCode = 301;
+        res.setHeader('Location', req.url + '/');
+        res.end('Redirecting to ' + req.url + '/');
+    }
+
+    send(req, req.url, {root: fixtures, etag: 'strong'})
+    .on('error', error)
+    .on('directory', redirect)
+    .pipe(res);
+    });
+    request(app)
+    .get('/name.txt')
+    .expect('etag', /^"[^"]+"$/)
+    .end(done);
+  })
+  
+  it('should not add an ETag header field', function(done){
+    var app = http.createServer(function(req, res){
+    function error(err) {
+        res.statusCode = err.status;
+        res.end(http.STATUS_CODES[err.status]);
+    }
+
+    function redirect() {
+        res.statusCode = 301;
+        res.setHeader('Location', req.url + '/');
+        res.end('Redirecting to ' + req.url + '/');
+    }
+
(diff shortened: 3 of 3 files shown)


dougwilson:
You cannot send back a strong ETag without being able to guarantee the bytes will be identical, i.e. you need to at least take a hash of the file contents.

In order for strong Etags to be an option in this module, you need to add this functionality. Please also ensure that it does not buffer the entire file contents into memory, as the files can be quite large.

README.md
```diff
@@ -59,7 +59,9 @@ the 4th byte in the stream.
 
 ##### etag
 
-Enable or disable etag generation, defaults to true.
+Accepts `'weak'`, `'strong'`, custom function `function(data){ return etag}`,
```
dougwilson: It is not clear what the "data" argument is in the documentation. Please elaborate in the document what this is and what it contains.

index.js
```diff
@@ -768,7 +768,12 @@ SendStream.prototype.setHeader = function setHeader(path, stat){
   }
 
   if (this._etag && !res.getHeader('ETag')) {
-    var val = etag(stat)
+    var val = null;
+    if(typeof this._etag == 'function') {
+        val = this._etag(stat);
```
dougwilson: Custom ETags are expected to be added by the headers event, which is the purpose of the surrounding if statement. If the intention is to move that functionality into a specific option, please remove the existing custom ETag functionality or at lease deprecate it.

test/send.js
```diff
@@ -102,6 +102,84 @@ describe('send(file).pipe(res)', function(){
...
+
+    function redirect() {
+        res.statusCode = 301;
+        res.setHeader('Location', req.url + '/');
+        res.end('Redirecting to ' + req.url + '/');
+    }
+
+    send(req, req.url, {root: fixtures, etag: 'strong'})
+    .on('error', error)
+    .on('directory', redirect)
```
dougwilson: Why is this event listener needed for this test?

test/send.js
```diff
@@ -102,6 +102,84 @@ describe('send(file).pipe(res)', function(){
...
+
+    function redirect() {
+        res.statusCode = 301;
+        res.setHeader('Location', req.url + '/');
+        res.end('Redirecting to ' + req.url + '/');
+    }
+
+    send(req, req.url, {root: fixtures, etag: false})
+    .on('error', error)
+    .on('directory', redirect)
```
dougwilson: Why is this event listener needed for this test?

test/send.js
```diff
@@ -102,6 +102,84 @@ describe('send(file).pipe(res)', function(){
...
+    }
+
+    function redirect() {
+        res.statusCode = 301;
+        res.setHeader('Location', req.url + '/');
+        res.end('Redirecting to ' + req.url + '/');
+    }
+
+    send(req, req.url, {root: fixtures, etag: 'strong'})
+    .on('error', error)
```
dougwilson: Why is this event listener needed for this test?

test/send.js
```diff
@@ -102,6 +102,84 @@ describe('send(file).pipe(res)', function(){
...
+    }
+
+    function redirect() {
+        res.statusCode = 301;
+        res.setHeader('Location', req.url + '/');
+        res.end('Redirecting to ' + req.url + '/');
+    }
+
+    send(req, req.url, {root: fixtures, etag: false})
+    .on('error', error)
```
dougwilson: Why is this event listener needed for this test?

test/send.js
```diff
@@ -102,6 +102,84 @@ describe('send(file).pipe(res)', function(){
...
+
+    function redirect() {
+        res.statusCode = 301;
+        res.setHeader('Location', req.url + '/');
+        res.end('Redirecting to ' + req.url + '/');
+    }
+
+    send(req, req.url, {root: fixtures, etag: function(){return 'test'}})
+    .on('error', error)
+    .on('directory', redirect)
```
dougwilson: Why is this event listener needed for this test?

test/send.js
```diff
@@ -102,6 +102,84 @@ describe('send(file).pipe(res)', function(){
...
+    }
+
+    function redirect() {
+        res.statusCode = 301;
+        res.setHeader('Location', req.url + '/');
+        res.end('Redirecting to ' + req.url + '/');
+    }
+
+    send(req, req.url, {root: fixtures, etag: function(){return 'test'}})
+    .on('error', error)
```
dougwilson: Why is this event listener needed for this test?

retep007:
I thought strong etags are properly generated in etag package, but as I see they are not, wouldn't it be better to add it there ?

dougwilson:
You're only giving the etag package a stat object, which does not contain enough information to generate a strong ETag.

If you want to add strong ETag support, here are the requirements:

1. The ETag must guarantee byte identical responses for the same ETag.
2. Do not read the entire file contents into memory at ant time.
3. Do not read the file contents from disk more than once.

dougwilson:
> Needed for expressjs/express#2936

I think you misunderstand the requirements for that change, because the change does not require any changes to be made to any dependencies.

retep007:
Yes, I saw your comment on it. Anyway I think strong etags should be added. So I will first finish expressjs/express#2936 and then take a look at this.

dougwilson:
> Anyway I think strong etags should be added.

I don't agree that it is even worth it. Not even nginx of Apache give you this for files, because a file can be 800MB, and computing a hash can take significant time and resources, providing a very simple path to DoS a server from many repeated file requests.

I'm going to close the pull request, on the grounds for the highlighted issues above. If you are interested in adding strong ETags here, please keep the requirements I highlighted above, and you may want to research pitfalls and the reasons other servers don't do strong ETags for files so you can't take into account all the prior research and implementations that have already been done.

PR #106, opened 2016-03-22
Added request in setHeaders.

rogermadsen:
Change-Id: I0d8395dbdaa431478e4e9db3149cab4e23fac181

### DIFF
--- a/index.js
+++ b/index.js
@@ -755,8 +755,9 @@ SendStream.prototype.type = function type(path) {
 
 SendStream.prototype.setHeader = function setHeader(path, stat){
   var res = this.res;
+  var req = this.req;
 
-  this.emit('headers', res, path, stat);
+  this.emit('headers', res, path, stat, req);
 
   if (!res.getHeader('Accept-Ranges')) res.setHeader('Accept-Ranges', 'bytes');
   if (!res.getHeader('Cache-Control')) res.setHeader('Cache-Control', 'public, max-age=' + Math.floor(this._maxage / 1000));


rogermadsen:
We have a use case where we need to set headers on the response based on headers on the request. The request object is however not available in the setHeaders function. With this change the request object is added as a last parameter (backwards compatible). Is this a change that you can accept? We would appreciate if you could accept it so we don't have to maintain a local copy of send. Thanks!

dougwilson:
It's your event listener always within a scope that has `req`? For example, the following works just fine as-is:
```js
var http = require('http');
var send = require('send');

var app = http.createServer(function(req, res){
  send(req, req.url)
  .on('headers', function (res) {
    res.setHeader('request-user-agent', req.headers['user-agent']);
  })
  .pipe(res);
}).listen(3000);
```

`req` is not included in the event on purpose, because the `req` object has nothing to do with the event. Does the above example not work for you?

rogermadsen:
Our use case is for serving static files with Express. Express is using npm package serve-static which is then using send. 

app.use(express.static('dist', { setHeaders: function (res, path, stat) {
   // We need request object here...
}));

dougwilson:
Ah. If you are using Express, just use `res.req`:
```js
app.use(express.static('dist', { setHeaders: function (res, path, stat) {
// We need request object here...
// it is available as res.req
}));
```

rogermadsen:
Didn't know that. Thanks! :-)

dougwilson:
No problem! Glad that worked out for you :)

PR #108, opened 2016-04-11
Add support for precompressed (for example gzip) content

gmokki:
Disabled by default (should it be enabled by default?).
There should be no overhead if disabled.
If `opts.precompressed` is set to true, tries to serve both .br and .gz files, but custom configuration (encoding to extension mappings) is also supported

If enabled then for every file that is served we look up each configured file extension (even if browser does not request compressed variants) and if any are found we set the `Vary` header. This is required for caching proxies that the clients my be using to work correctly.

If matching file between `Accept-Encoding` and enabled precompression formats is found then the compressed file contents are served with `Content-Encoding` header. If multiple supported matches are found the first configured in the preferences will be used.


dougwilson:
Thanks, @gmokki! I have a question regarding the added functionality:

If I were to enable `precompressed` with this code and I have a folder full of `*.tar.gz` files, how can I prevent the clients from accessing the files as `*.tar` (as this will cause the client to automatically ungzip potentially giant tarballs)?

gmokki:
I do not think there can be anything to prevent that (at least with this patch). It works the same way as apache/nginx etc when configured with support for precompressed resources.
I'd say the main defence is that one does not link to `.tar` files, but directly to `.tar.gz` files. If you directly refer to the `.gz` file the patch does not set the content-encoding headers and thus does not cause the client to decompress the content.

gmokki:
More concrete answer:
Assuming you have files `file.tar` and `file.tar.gz` in folder where the uncompressed file.tar.gz contents matchs exactly to to the file.tar.

Before this patch:
- When requesting `file.tar.gz` you get the exact (compressed) file.
   - the `Accept-Encoding` header with or without gzip has no effect
- When requesting `file.tar` you get the exact (uncompressed) file.
   - the `Accept-Encoding` header with or without gzip has no effect

After this patch:
- When requesting `file.tar.gz` you get the exact (compressed) file.
   - the `Accept-Encoding` header with or without gzip has no effect
- When requesting `file.tar`
   - without `Accept-Encoding: gzip` header you get the exact (uncompressed) file
   - with `Accept-Encoding: gzip` header you get the `tar.gz` file with `Content-Encoding: gzip` response
      - the only difference is that the transferred wire data is smaller and the http client transparently decompresses to the `file.tar` as requested

For applications using the `precompressed` feature they just need to make sure that the contents of the uncompressed and compressed files match. As long as that holds no difference can be observed by the compilant http client users.

dougwilson:
Ah, @gmokki, gotcha. Yea, I looked at the nginx source and what they do is just unconditionally add `.gz` to the file name when the client wants to get the gzip encoding, and there are no other real checks. I think this is probably OK, but I think the PR needs a lot more documentation fleshed out.

gmokki:
Can you give a hint of what kind of extra documentation in addition to the short README.md documentation you would like to have.

dougwilson:
Hi @gmokki, currently the README change in this pull request is only the following:

> Enable or disable serving of precompressed content, defaults to `false`.
> If set to `true` checks for brotli and gzip. An example value to serve
> other formats: `[{encoding: 'bzip2', extension: '.bz2'}]`

There are a lot of questions that raises, including how to even use it. What values can I use for `encoding`? What about `extension`? How does this module determine which file to serve? How do I layout the files on the disk? How can we avoid questions like the one I had above?

I'm not asking you these questions, they are just what comes to mind when reading the README. We'll need enough information in the README for someone to use the feature. Consider that you may take it for granted what "precompressed" means and how to lay it out, but many people have never done that before.

gmokki:
I added the details you requested into the documentation with much longer clarifications. I also added a few generic tips on offline compression.

index.js
```diff
@@ -285,6 +295,24 @@ SendStream.prototype.isConditionalGET = function(){
...
+ * @return {Array}
+ * @api private
+ */
+
+SendStream.prototype.getAcceptEncodingExtensions = function() {
+  var accepted = []
+  var header = this.req.headers['accept-encoding']
+  if (header) {
+    this._precompressionFormats.forEach(function (format) {
+      if (header.indexOf(format.encoding) >= 0) accepted.push(format.extension)
```
dougwilson: Looks like this is broken for web browsers that send "no-gzip, deflate", which was common for a time (Apache, and I assume others, understand no-gzip).

This also does not seem to take into account the quality parameters at all, which is very common from CDNs, as they usually want a specific encoding to store.

dougwilson:
Thanks! I just realized I still have not actually looked at the implementation, so doing a little bit now.

index.js
```diff
@@ -637,10 +666,47 @@ SendStream.prototype.sendFile = function sendFile(path) {
...
+      debug('stat "%s%s"', p, format.extension);
+      fs.stat(p + format.extension, function onstat(err, contentStat) {
+        if (!err) state.contents.push({ext: format.extension, encoding: format.encoding, contentStat: contentStat})
+        if (--state.extensionsToCheck == 0) sendPreferredContent(p, stat, state.contents)
+      })
+    })
+  }
+
+  function sendPreferredContent(p, stat, contents) {
+    if (contents.length) self.res.setHeader('Vary', 'Accept-Encoding')
```
dougwilson: This will overwrite any other Vary already set. Let's append this to the field name if already set instead of overwriting.

index.js
```diff
@@ -637,10 +666,47 @@ SendStream.prototype.sendFile = function sendFile(path) {
...
+    if (!self._precompressionFormats) return self.send(p, stat)
+
+    var state = {
+      contents: [],
+      extensionsToCheck: self._precompressionFormats.length
+    }
+
+    self._precompressionFormats.forEach(function (format) {
+      debug('stat "%s%s"', p, format.extension);
+      fs.stat(p + format.extension, function onstat(err, contentStat) {
```
dougwilson: Why bother stating files we don't intend to send to the client? For example, this will stat for .gz and .br, even if the client only accepts gzip.

README.md
```diff
@@ -85,6 +85,43 @@ Provide a max-age in milliseconds for http caching, defaults to 0.
...
 
+##### precompressed
+
+Precompressed files are extra static files that are compressed before
+they are requested, as opposed to compressing on the fly. Compressing
+files once offline (for example during site build) allows using
+stronger compression methods and both reduces latency and lowers cpu
+usage when serving files.
+
+The 'precompressed` option enables or disables serving of precompressed
```
dougwilson: Looks like a typo in the formatting here.

index.js
```diff
@@ -114,6 +114,16 @@ function SendStream(req, path, options) {
     ? normalizeList(opts.extensions, 'extensions option')
     : []
 
+  if (Array.isArray(opts.precompressed)) {
+    this._precompressionFormats = opts.precompressed
+  } else if (opts.precompressed) {
+    this._precompressionFormats = [{encoding: 'br', extension:'.br'}, {encoding:'gzip', extension:'.gz'}]
```
dougwilson: If, by default, we are going to search for files with the `.br` extension, I want to be sure this is actually the official file extension for this format and it does not become something else at some point. Can you point me to somewhere that says the file extension for brotli (ideally some place with authority in the brotli project). All I can find is the `Content-Encoding` defined as `br`, but no mention of a file extension.

index.js
```diff
@@ -637,10 +666,47 @@ SendStream.prototype.sendFile = function sendFile(path) {
...
       if (stat.isDirectory()) return next()
-      self.emit('file', p, stat)
-      self.send(p, stat)
+      checkPrecompressionAndSendFile(p, stat)
     })
   }
+
+  function checkPrecompressionAndSendFile(p, stat) {
+    self.emit('file', p, stat)
+    if (!self._precompressionFormats) return self.send(p, stat)
```
dougwilson: `{ precompressed: [] }` causes all responses to hang.

index.js
gmokki: https://github.com/google/brotli/issues/299#issuecomment-173892279

dougwilson: Content-encoding is not a file extension, otherwise it would be .gzip, not .gz. Can you link me to the place where the file extension is specified?

hacksparrow:
This is a good feature to have. @gmokki can you can address the issues pointed out by @dougwilson?

gmokki:
Yes, I will definitely try to address all pointed issues in next few days. Thank you for the through review and comments.

dougwilson:
Thanks, @gmokki! It sounds like this won't make it for the Express.js 4.14 release cut tonight, then. It'll probably be a few months before the next minor, and this module is locked to Express.js release schedules, so you'll have plenty of time to address!

index.js
gmokki: The comment I linked to has link to draft specification (https://github.com/madler/brotli/blob/master/br-format-v3.txt) for the .br file format. It gives an example of .tar.br as a file name.

gmokki:
I just fixed all the issues that @dougwilson pointed out.

I have already added a encoding quality parsing function to tomcat (https://github.com/apache/tomcat/blob/trunk/java/org/apache/catalina/servlets/DefaultServlet.java#L1096) - I'll try to add it now to here too.
But if that feature does not make it in time would it be possible to merge this without the encoding quality parsing? None of the common browsers send quality values. Actually so far all web servers that I have patched with brotli support just assumed that gzip;q=0 meant that gzip was supported.

dougwilson:
Thanks, @gmokki! It looks like there is still the outstanding stat issue (unless you fixed it in one of the commits?). The quality issue is a real issue, at least from CDNs. I maintain the `compression` module, which does similar negotiation, and I definitely know people have had real issues with this, and since this serves static files, which are pretty much the main thing to put behind a CDN, then it is an issue here as well, and I would rather get this done right than trying to rush this & make mistakes.

hacksparrow:
+1 for not rushing it.

index.js
dougwilson: Cool, sorry @gmokki, I did not see the link in the linked to comment. Reading through that spec, it sounds like the file framing for the `.br` file format is not the same as the framing used in the `br` Content-Encoding, is that correct? If so, we should not be using that file extension to serve up as if it were the content encoding (I am comparing the content encoding https://tools.ietf.org/html/draft-alakuijala-brotli-11#section-2 to the referenced file framing https://github.com/madler/brotli/blob/master/br-format-v3.txt).

Do you know if we send out a brotli stream that has the file framing around it as `Content-Encoding: br` if the clients will read it? I would imagine that the decoders will choke on the first 4 bytes of the file frame if they are trying to read it as a content stream, no?

If you can try it, can you see what a web browser does when `Content-Encoding` is `br` and the following bytes are the body: `ce b2 cf 81 84 06 00 27`?

index.js
gmokki: The reason to stat for files that client does not accept: proxies.
Lets assume that our server is behind a caching proxy.
First client sends request without accept-encoding (aka uncompressed). If we do not stat for all the files on the server side even in that case the server will not set the Vary header.
And when a second client comes in with accept-encoding header the proxy that did not see a Vary header will serve the cached uncompressed content.

Thus the downside of precompressed option is that you need to stat for files even for clients that do not request them.

gmokki:
I just added code for the accept-encoding quality handling.

I just added comment on why the stats are needed always when precompressed option is enabled. If that were not the case the precompression option could default to on.

I think I have now really addressed all comments. Do provide more feedback on bugs/missing features and I'll try to address them too.

index.js
dougwilson: Hi @gmokki, I'm not saying to not stat at all. I'm saying why stat for all of them. In your example:

> First client sends request without accept-encoding (aka uncompressed).

You don't need to stat at all in this case. Always send the `Vary` header, because the file system can change at any time and a `.gz` file can later appear when it was not there before (think if someone combines this with background compressing on-demand). If you don't expect the file system to change, then you would be caching all these looks, but you're not.

dougwilson:
Thanks, @gmokki! In order to get the 4.14 release ready for tonight and our TC meeting, I cannot look any more at this PR for now. I'll take another look tomorrow after 4.14 for sure. There is still the comment regarding the `.br` file framing vs the stream if you can take a look into that as well!

dougwilson:
Also, the CI tests are still taking their sweet time to run (they still haven't gotten to your pushes from earlier today), so we have to at least wait on those to consider merging.

Also, real quick, I do have a couple more thoughts on the header stuff: have you considered using the modules we have already owned by the Express TC to do some of that heavy lifting? Did they not work out (and would it be possible for us to modify them so they can be used here)? Most of them can be found in https://github.com/jshttp (our organizations are https://github.com/expressjs , https://github.com/pillarjs and https://github.com/jshttp).

gmokki:
I'm not an authority on the brotli file formats nor their official extension.
For the extension I can say that all other web servers I know have the brotli static file serving support (ngingx, tomcat, jetty) and apache examples all refer to extension being .br.
The file format is outside of the scope of this module. It is the users responsibility to make sure that file ending with .gz/.br can be decompressed by the browser.

I originally tried to avoid extra dependencies. But I could use https://github.com/jshttp/vary to modify the vary header and https://github.com/jshttp/negotiator to parse the preferred accept-encoding header. I'll try to modify the code to use them tomorrow. Now that the code has tests for both cases taking them into use should be simple.

dougwilson:
> For the extension I can say that all other web servers I know have the brotli static file serving support (ngingx, tomcat, jetty) and apache examples all refer to extension being .br.

That doesn't make it the correct thing to do. For example, I could have just accepted this and been in the same boat. Just because others do something doesn't automatically make it correct. In order to accept this PR as-is, with `.br` being check _by default_, it has to make sense, and as such, I need you to perform the investigation I asked about the framing, otherwise it will have to wait until I have time to do that depth of digging, which I will do in the order I have received PRs to be fair to all other people who have open PRs. This PR is currently `#82` in the line...

index.js
gmokki: Always sending Vary means that it is also sent for uncompressable files such as images. Which means that caching proxies have to keep about 5-10 (the number of unique accept-encoding headers used by common browsers) versions of each image in cache instead of only one.

If served files are changed while server is running then the browser/proxy caches will in any case have stale data and missing Vary header for a file that was cached before it had precompressed variants is correct behavior.

I'm not familiar with good caching techniques in nodejs, but in any case I think caching cannot be enabled by default. It will cause wrong data to be sent out (Content-Length/Vary) if the files are changed while server is running - harming both development and production environments.

Maybe I (we?) should benchmark how much performance is lost by the extra stats?

index.js
gmokki: I tried the byte sequence - both firefox and chrome report it as content encoding error.

gmokki:
@dougwilson: I think you are correct that adding explicit brotli support is premature. The file format is not yet finalized. I just pushed a new version that removes brotli from the default set of formats and updates the documentation to match. 

I also tried using the https://github.com/jshttp/accepts for choosing the best encoding format. Unfortunately that does not yet work because the module does not honor the server preference order for values that have equal quality level in the client request.

index.js
dougwilson: Ah, that's what I was afraid of after reading the specs. It's possible they may remove the framing at some point, but currently it is in there, and it would probably be really confusing to users how to make a `.br` file that this module can use, since eventually tools will exist to make `.br` files and they will contain the file framing.

index.js
dougwilson: > Always sending Vary means that it is also sent for uncompressable files such as images. Which means that caching proxies have to keep about 5-10 (the number of unique accept-encoding headers used by common browsers) versions of each image in cache instead of only one.

Maybe a solution to this can be to allow the user to configure which types are actually compressed (similar to how the `compression` middleware works).

> If served files are changed while server is running then the browser/proxy caches will in any case have stale data and missing Vary header for a file that was cached before it had precompressed variants is correct behavior.

Sure, though by default, this module sends out `max-age=0` unless changed. Currently we have no issue with allowing files to change under this module, and there have been various issues relating to improving this over the years. As far as this module is concerned, any file can be changed at any time, this would include we should assume that a file that did not have a corresponding `.gz` file at one point in time may have one generated later, and should signal such in the response, i.m.o

> Maybe I (we?) should benchmark how much performance is lost by the extra stats?

Yes, please benchmark. Remember that Node.js absolutely sucks for any file operations, since all file operations (including stats) are not async, but rather executed sync in a background thread pool with a limited number of threads (defaults to 4). As such, influx in any file operations can really hurt Node.js-based applications.

dougwilson:
> I also tried using the https://github.com/jshttp/accepts for choosing the best encoding format. Unfortunately that does not yet work because the module does not honor the server preference order for values that have equal quality level in the client request.

Is it possible to get that added? I think I remember seeing that wasn't in there when dealing with `compression` issues, so it would be nice if we can get that there to keep us from continuing to punt on the issue :)

gmokki:
My negotiator pull request https://github.com/jshttp/negotiator/pull/49 is still not merged, but I updated this pull request to use it. I also did some worst case benchmarking: when enabling the precompressed check with one extra stat the performance goes down at most by 20% when serving 1 byte files to localhost with single infinite depth http pipeline connection. I doubt that it is observable in actual use cases.

gmokki:
Rebased to fix merge conflicts. Unfortunately the negotatior PR is still pending.

DylanPiercey:
Is there anything that could be done to aid in landing this PR?

PR #113, opened 2016-08-02
Add support for multipart/byterange responses

jcready:
Properly handles byte-range requests with multiple non-overlapping ranges by responding with a multipart/byterange response. For details on the specification for handling this type of request see [RFC 7233](https://tools.ietf.org/html/rfc7233#section-4.1)

For example, a request looking like this:
```
GET /nums
Host: example.com
Range: bytes=0-1,3-3,5-6,7-8
```

Would generate a response like this:

```
HTTP/1.1 206 Partial Content
Accept-Ranges: bytes
ETag: W/"39aa72-153af0d74c0"
Content-Type: multipart/byteranges; boundary=BYTERANGE_1470163149308
Content-Disposition: attachment; filename="nums"
Date: Tue, 02 Aug 2016 18:39:09 GMT
Transfer-Encoding: chunked
Connection: close

--BYTERANGE_1470163149308
Content-Type: application/octet-stream
Content-Range: bytes 0-1/9

12
--BYTERANGE_1470163149308
Content-Type: application/octet-stream
Content-Range: bytes 3-3/9

4
--BYTERANGE_1470163149308
Content-Type: application/octet-stream
Content-Range: bytes 5-8/9

6789
--BYTERANGE_1470163149308--
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -281,6 +281,9 @@ SendStream.prototype.error = function error (status, error) {
   var res = this.res
   var msg = statuses[status]
 
+  // do not set headers if they have already been sent
+  if (res._headerSent) return res.end(msg)
+
   // clear existing headers
   clearHeaders(res)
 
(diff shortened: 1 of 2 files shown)


dougwilson:
Hi @jcready, sorry I haven't gotten around to looking at your PRs yet :) I see you closed this and opened #114 as an updated version of this PR, so I will make any comments over there (and linking them together for future reference).

jcready:
@dougwilson no worries. I did close this one in favor of #114 since I had so many issues with getting the tests to pass in old versions of node on travis. Even using `nvm` to test old versions didn't seem to ensure they would also pass in travis (possible race condition issues when trying to simulate file stream errors).

PR #114, opened 2016-08-04
Add support for multipart/byterange responses

jcready:
Properly handles byte-range requests with multiple non-overlapping ranges by responding with a multipart/byterange response. For details on the specification for handling this type of request see [RFC 7233](https://tools.ietf.org/html/rfc7233#section-4.1)

For example, a request looking like this:

```
GET /nums
Host: example.com
Range: bytes=0-1,3-3,5-6,7-8
```

Would generate a response like this:

```
HTTP/1.1 206 Partial Content
Accept-Ranges: bytes
ETag: W/"39aa72-153af0d74c0"
Content-Type: multipart/byteranges; boundary=BYTERANGE_IS5K8FWO
Content-Disposition: attachment; filename="nums"
Date: Tue, 02 Aug 2016 18:39:09 GMT
Content-Length: 311
Connection: close

--BYTERANGE_IS5K8FWO
Content-Type: application/octet-stream
Content-Range: bytes 0-1/9

12
--BYTERANGE_IS5K8FWO
Content-Type: application/octet-stream
Content-Range: bytes 3-3/9

4
--BYTERANGE_IS5K8FWO
Content-Type: application/octet-stream
Content-Range: bytes 5-8/9

6789
--BYTERANGE_IS5K8FWO--
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -823,7 +823,7 @@ SendStream.prototype.streamMultipart = function streamMultipart (path, options) 
     // iterate through all the ranges
     asyncSeries(options.ranges, function (range, idx, next) {
       if (finished) return next()
-      var isLast = idx >= range.length - 1
+      var isLast = idx >= options.ranges.length - 1
       var partHeaders = (
         (idx ? CRLF : '') +
         '--' + BOUNDARY + CRLF +
@@ -1065,7 +1065,7 @@ function asyncSeries (array, iteratee, done) {
 /**
  * Only allow a function to run once
  *
- * @param {fn} function
+ * @param {function} fn
  * @private
  */
 
--- a/index.js
+++ b/index.js
@@ -36,6 +36,7 @@ var util = require('util')
  * @private
  */
 
+var basename = path.basename
 var extname = path.extname
 var join = path.join
 var normalize = path.normalize
@@ -70,6 +71,12 @@ var UP_PATH_REGEXP = /(?:^|[\\\/])\.\.(?:[\\\/]|$)/
 
 module.exports = send
 module.exports.mime = mime
+
+/**
+ * Is the node version is < 0.10
+ */
+
+var isLegacyNodeVersion = Boolean(EventEmitter.listenerCount)
 
 /**
  * Shim EventEmitter.listenerCount for node.js < 0.10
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -771,6 +800,71 @@ SendStream.prototype.stream = function stream (path, options) {
...
+        'Content-Range: ' + contentRange('bytes', fileSize, range) + CRLF +
+        CRLF
+      )
+      /* istanbul ignore next */
+      range.fd = isLegacyNodeVersion && idx ? null : fd
+      range.autoClose = isLast
+      stream = fs.createReadStream(path, range)
+      stream.on('error', next)
+      stream.on('end', function onpartend () { process.nextTick(next) })
+      if (!idx) self.emit('stream', stream)
```
jcready: Not sure how you want to handle this. I'm creating a new read stream for every range/part so right now I'm only emitting the 'stream' event on the first iteration. I could emit this event on every iteration, but I feel like that might be confusing for a developer using this library since they would end up seeing this event get triggered N times, but the 'end' event would only get emitted once.

index.js
```diff
@@ -620,17 +659,25 @@ SendStream.prototype.send = function send (path, stat) {
...
 
-      // adjust for requested range
-      offset += ranges[0].start
-      len = ranges[0].end - ranges[0].start + 1
+      if (ranges.length === 1) {
+        res.setHeader('Content-Range', contentRange('bytes', len, ranges[0]))
+        // adjust for requested range
+        offset += ranges[0].start
+        len = ranges[0].end - ranges[0].start + 1
+      } else if (ranges.length > 1) {
```
jcready: Maybe this path should be restricted to some maximum array length. It could be made configurable but would default to say 10.

jcready:
@dougwilson I'd greatly appreciate a review on this :) I promise I won't make any more commits to this branch. I think I finically got everything nailed down with regards to node < 0.10.

jcready:
Closing this because I'm working on something better. Will get a new PR up soon.

dougwilson:
Sweet. I'm sorry I haven't really sat down to review it, but I have definitely been opening the emails when you make commits, which has mainly been on the race condition PR recently.

dougwilson:
I did honestly kind of drop this a bit since I didn't assign it to myself, it wasn't showing up in the GitHub search results I try to use to track everything happening.

develar:
This feature is still not supported? Doesn't work for me :(

daveying:
This will solve my problem #165 , really happy to know this is in progress.

daveying:
I am definitely willing to do that :smile: , but I noticed @jcready already have some work done on this issue.

PR #116, opened 2016-08-15
Accept a `Function` object as `index` option

piranna:


piranna:
This is a streamlined from scratch reimplementation of issue #89 on top of `master` code to be as minimal and focused as possible.

PR #124, opened 2016-12-12
Prevent safari strict mode error

dereke:
This is a fix for this issue https://github.com/pillarjs/send/issues/80

> SyntaxError: Cannot declare a parameter named 'error' as it shadows
the name of a strict mode function.

I've added a fairly descriptive test for this so that if the functionality is reversed then the test will catch it.

Why bother doing this?

I am running express in the browser with a virtual http layer between my http client and the server. This allows me to test a single page web application against the actual server without having to listen on any ports or pay any overhead of making an actual http request. This is enabled by https://www.npmjs.com/package/vinehill and I am testing my app with https://www.npmjs.com/package/browser-monkey - if you are interested.

### DIFF
--- a/index.js
+++ b/index.js
@@ -269,7 +269,7 @@ SendStream.prototype.maxage = deprecate.function(function maxage (maxAge) {
  * @private
  */
 
-SendStream.prototype.error = function error (status, error) {
+SendStream.prototype.error = function (status, error) {
   // emit if listeners instead of responding
   if (listenerCount(this, 'error') !== 0) {
     return this.emit('error', createError(error, status, {
--- a/test/send.js
+++ b/test/send.js
@@ -1278,6 +1278,14 @@ describe('send.mime', function () {
   })
 })
 
+describe('safari compatibility', function () {
+  describe('.error function', function () {
+    it('does not used named function which causes strict mode error', function () {
+      assert.equal(new send().error.name, '')
+    })
+  })
+})
+
 function createServer (opts, fn) {
   return http.createServer(function onRequest (req, res) {
     try {


dougwilson:
Thanks for the effort, but this is a wontfix. If you need Safari has admitted this is a bug in their JS and fixed it. If you need compatibility, run the source through a transformation process.

dougwilson:
If you truly want this module yo work first class in other environments than Node.js, please contribute a CI setup to run this in Safari and others, then we can consider this fix, which is specific to an environment we do not support at this point.

From the last PR for this:

> Now, perhaps the suggested course of action is to get this module to function in web browsers like Safari, contribute CI setup to test this module in Safari and other browsers, and bring web browsers to be supported by this module, if you are up to the task :)

dereke:
I hadn't thought about transforming the code, that is a pretty blunt instrument but it did the trick.

thanks for the quick response.

dougwilson:
Hi @dereke no problem. Usually people have to run this through a transformation just to get it up & into the web browser at all (since it is CJS and using Node.js dependencies), so having one to work-around Safari bugs seems logical (to me). You never know when this module or a dependency will suddenly become ES6 or such since that works in Node.js without issue, but I have no idea how well that works in Safari. I'm sure trying to load things in environments they were not expecting will run into weird behavior every so often :)

dereke:
yes browser bugs are all too common but at least nothing like the past.  I was already using browserify so it was really no big deal to just transform the code in the end.

I must say I was somewhat shocked when I was able to get `require('express')` working, it required a little [monkey patching of http](https://github.com/dereke/vinehill/blob/01f946193b96cbed69a9d846cc67efe44dd8df7b/index.js#L5) but was really not an issue. Thanks for all your work on these libs.

dougwilson:
No problem :) We do try and keep it simplistic.

PR #127, opened 2017-01-14
Fix response headers usage with upcoming node changes

mscdex:


index.js
```diff
@@ -405,7 +405,22 @@ SendStream.prototype.onStatError = function onStatError (error) {
  */
 
 SendStream.prototype.isFresh = function isFresh () {
-  return fresh(this.req.headers, this.res._headers)
+  var resHeaders = this.res._headers || {}
+
+  // Support changes in node with how response headers are stored
+  var headerNames = this.res._headerNames || {}
+  var keys = Object.keys(resHeaders)
+  if (keys.length > 0 && headerNames[keys[0]] === undefined) {
```
dougwilson: I think I understand what this check is attempting to do, but the popular https://github.com/molnarg/node-http2 module does not emulate the `_headerNames` property, which I assume this check then would assume that `_headers` object is the new style rather than the old style, is that correct?

dougwilson: I suppose that in that case, this would just end up with all the values being `undefined` (or the second copy of the header if someone set the value to an array).

dougwilson:
Since I never heard anything back from my question / thoughts above, I took a closer look and (1) added a build in Travis CI against Node.js 8 nightly and then (2) simplified the implementation that kept the library compatible with `http2` and the like.

PR #131, opened 2017-02-12
Support `If-Match` and `If-Unmodified-Since`

### LINKED ISSUE #130: Support If-Match and If-Unmodified-Since
### [`If-Match`](https://tools.ietf.org/html/rfc7232#section-3.1)

The server will send back the requested resource only if it matches one of the listed `ETags` otherwise the response will be a `412` (Precondition Failed) error.

> The "If-Match" header field makes the request method conditional on the recipient origin server either having at least one current representation of the target resource, when the field-value is "*", or having a current representation of the target resource that has an entity-tag matching a member of the list of entity-tags provided in the field-value.

### [`If-Unmodified-Since`](https://tools.ietf.org/html/rfc7232#section-3.4)

The server will send back the requested resource only if it has not been last modified after the given date. If the request has been modified after the given date, the response will be a `412` (Precondition Failed) error.

> The "If-Unmodified-Since" header field makes the request method conditional on the selected representation's last modification date being earlier than or equal to the date provided in the field-value. This field accomplishes the same purpose as If-Match for cases where the user agent does not have an entity-tag for the representation.
>
> A recipient MUST ignore If-Unmodified-Since if the request contains an If-Match header field; the condition in If-Match is considered to be a more accurate replacement for the condition in If-Unmodified-Since, and the two are only combined for the sake of interoperating with older intermediaries that might not implement If-Match.

jcready:
Resolves #130

### DIFF
--- a/index.js
+++ b/index.js
@@ -314,7 +314,7 @@ SendStream.prototype.hasTrailingSlash = function hasTrailingSlash () {
  * @api private
  */
 
-SendStream.prototype.isSatisfyingPreconditions = function isSatisfyingPreconditions (stat) {
+SendStream.prototype.isSatisfyingPreconditions = function isSatisfyingPreconditions () {
   // if-match
   var match = this.req.headers['if-match']
 
--- a/index.js
+++ b/index.js
@@ -308,6 +308,39 @@ SendStream.prototype.hasTrailingSlash = function hasTrailingSlash () {
 }
 
 /**
+ * Returns `true` if any request preconditions are failing
+ *
+ * @return {Boolean}
+ * @api private
+ */
+
+SendStream.prototype.isFailingPrecondition = function isFailingPrecondition (stat) {
+  // if-match
+  var match = this.req.headers['if-match']
+  var etag = this.res._headers['etag']
+  var strongEtag = /^(?!W\/)/
+
+  if (match) {
+    if (strongEtag.test(etag)) {
+      return !match.split(/ *, */).some(function (match) {
+        return match === '*' || match === etag
+      })
+    }
+  }
+
+  // if-unmodified-since
+  var unmodifiedSince = Date.parse(this.req.headers['if-unmodified-since'])
+  var lastModified = Date.parse(this.res._headers['last-modified'])
+  if (isNaN(lastModified)) lastModified = stat.mtime.getTime()
+
+  if (!isNaN(unmodifiedSince) && !isNaN(lastModified)) {
+    return unmodifiedSince <= lastModified
+  }
+
+  return false
+}
+
+/**
  * Check if this is a conditional GET request.
  *
  * @return {Boolean}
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -308,15 +308,50 @@ SendStream.prototype.hasTrailingSlash = function hasTrailingSlash () {
...
+ */
+
+SendStream.prototype.isFailingPrecondition = function isFailingPrecondition (stat) {
+  // if-match
+  var match = this.req.headers['if-match']
+  var etag = this.res._headers['etag']
+  var strongEtag = /^(?!W\/)/
+
+  if (match) {
+    if (strongEtag.test(etag)) {
```
dougwilson: From my reading of RFC 7232, `If-Match: "foo"` should fail with 412 if this module is about to respond with `ETag: W/"bar"`, but this will not fail here since it skips matching if this module is about to send a weak ETag, which seems to not match the requirements.

index.js
```diff
@@ -308,15 +308,50 @@ SendStream.prototype.hasTrailingSlash = function hasTrailingSlash () {
...
+      return !match.split(/ *, */).some(function (match) {
+        return match === '*' || match === etag
+      })
+    }
```
dougwilson: It also seems to be missing on the way `*` works when this module is about to send a 404. It seems to me that if the request says `If-Match: "foo"`, then this module MUST 412 on the response if it would have normally 404ed from:

> If the field-value is "*", the condition is false if the origin server does not have a current representation for the target resource.

index.js
```diff
@@ -308,15 +308,50 @@ SendStream.prototype.hasTrailingSlash = function hasTrailingSlash () {
...
+ * Returns `true` if any request preconditions are failing
+ *
+ * @return {Boolean}
+ * @api private
+ */
+
+SendStream.prototype.isFailingPrecondition = function isFailingPrecondition (stat) {
+  // if-match
+  var match = this.req.headers['if-match']
+  var etag = this.res._headers['etag']
```
dougwilson: Brought up in #127, we have to remove all uses of `res._headers` property, so best not to add new ones, at least :)

index.js
```diff
@@ -308,15 +308,50 @@ SendStream.prototype.hasTrailingSlash = function hasTrailingSlash () {
...
+    if (strongEtag.test(etag)) {
+      return !match.split(/ *, */).some(function (match) {
+        return match === '*' || match === etag
+      })
+    }
+  }
+
+  // if-unmodified-since
+  var unmodifiedSince = Date.parse(this.req.headers['if-unmodified-since'])
+  var lastModified = Date.parse(this.res._headers['last-modified'])
```
dougwilson: Brought up in #127, we have to remove all uses of `res._headers` property, so best not to add new ones, at least :)

dougwilson:
This will land in 0.15.0

dougwilson:
Unless you strongly disagree, I'll make the following alterations upon landing:

* Hoist the regular expression to a variable
* Set the `If-Match` back to a weak etag match; I know we discussed, but the `If-Range is also using the wrong match and probably makes sense for now until we get strong tags. Technically the modified is weak anyway, and the etag we make is slightly stronger :shrugs:

You're welcome to make the changes in the PR if I haven't landed this yet, of course, just don't want to make it feel like it's in limbo.

jcready:
@dougwilson sounds good to me.

dougwilson:
👍 I'm getting closer on finishing the merge / changes

PR #133, opened 2017-02-26
Renamed error param to err; to fix Error in strict

gxela:
SyntaxError: Cannot declare a parameter named 'error' in strict mode

Getting this error when running in phantomJS. Making this change fixes the SyntaxError and makes it go away!

### DIFF
--- a/index.js
+++ b/index.js
@@ -276,10 +276,10 @@ SendStream.prototype.maxage = deprecate.function(function maxage (maxAge) {
  * @private
  */
 
-SendStream.prototype.error = function error (status, error) {
+SendStream.prototype.error = function error (status, err) {
   // emit if listeners instead of responding
   if (listenerCount(this, 'error') !== 0) {
-    return this.emit('error', createError(status, error, {
+    return this.emit('error', createError(status, err, {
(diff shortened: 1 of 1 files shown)


gxela:
screen shot 2017-02-26 at 10 15 49 am

gxela:
Perhaps, the reason why is, strict mode is confused about the param being the same name as the function name

dougwilson:
This is a bug in only a few versions of Webkit. Please see the past #124 and #80

dougwilson:
The summary from the previous discussions are three different possibilities:

1. We can make this change here, but we don't make them without tests. Since this is a Safari bug, this means we need someone to come and help us make this module work in Safari in the first place and get a CI setup to ensure tat it stays working.
2. Depending on what you're attempting to do, perhaps you don't even need to load this module in Safari, as I doubt this module would work at all, so there is no reason to waste loading it up.
3. If you really need it, just pass it though a source transformer like browserify to make Node.js module compatible in the web browser (otherwise I don't think it's possible for this module to work anyhow).

gxela:
Thanks for this!

PR #135, opened 2017-03-04
add option autoEnd to enable append customized data after pipe

xdcarter:
add option autoEnd to give chance to append some customized data to the raw file。in the case of package tool, I want to auto add the module wrapper code to the raw JS in dev mode，to avoid the trouble for the package building。

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,13 @@ Serve files relative to `path`.
 Byte offset at which the stream starts, defaults to 0. The start is inclusive,
 meaning `start: 2` will include the 3rd byte in the stream.
 
+##### autoEnd
+
+Enable or disable ending the stream of the http response  after pipe, defaults to true.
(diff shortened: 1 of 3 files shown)


dougwilson:
Thanks for adding docs! I didn't really look over it at the time, but I saw you added the docs, and went to look. It made we wonder why would you add the `autoEnd` option to the main options instead of accepting the option through our `.pipe()` method, especially since it looks like the intention is to mimic the `.pipe()` option from Node.js core? Seems like updating the PR to just be `.pipe(res, { end: false })` to match `.pipe()` from core streams would make the most sense.

xdcarter:
Cuz I cannot add my new commit to the existed pull request #135, I have to close #135 and open new PR for your review. At this moment, I have no change request to make. Would you pls review #137 and see if it can be merged to master?

PR #140, opened 2017-04-29
Add support for Cache-Control: immutable

jcready:
If the `maxAge` option is set to `Infinity` then the resulting header will be set:

```
Cache-Control: public, max-age=31536000, immutable
```

Resolves https://github.com/expressjs/express/issues/3197

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,11 @@ system's last modified value.
 
 Provide a max-age in milliseconds for http caching, defaults to 0.
 This can also be a string accepted by the
-[ms](https://www.npmjs.org/package/ms#readme) module.
+[ms](https://www.npmjs.org/package/ms#readme) module. The maximum max-age
+is 1 year or 31536000000ms. If set to `Infinity` the [`immutable` Cache-Control
+directive](https://bitsup.blogspot.de/2016/05/cache-control-immutable.html) will
+be added (e.g. `Cache-Control: public, max-age=31536000, immutable`).
+
 
 ##### root
 
--- a/index.js
+++ b/index.js
@@ -165,6 +165,7 @@ function SendStream (req, path, options) {
   this._maxage = typeof this._maxage === 'string'
     ? ms(this._maxage)
     : Number(this._maxage)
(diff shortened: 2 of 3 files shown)


dougwilson:
So after some (light) reading, I realize (and I don't think I did before) that "immutable" is actually useful for any cache duration, so saying max age 4 hours + immutable is a valid configuration. I am going to land this still, still with full credit to @jcready , but with the following API change: instead of `{ maxAge: Infinity }` = add immutable flag, a new `immutable` option will be added, with the default as `false`, such that `{ immutable: true }` = immutable (and you can do `{ maxAge: '4h, immutable: true }`).

dougwilson:
Another option, just to throw it out there (though it's high visibility), is to actually _deprecate_ the `maxAge` option and have `cacheControl` accept an object so you can `{ cacheControl: { maxAge: '4h' } }` or `{ cacheControl: { maxAge: '6h', immutable: true } }` or some other combination.

I'm not 100% sure how I feel about deprecating the `maxAge` option since it's been here for so long. Perhaps deliver the separate `{ immutable: true }` option now and in a later release deprecate both `maxAge` and `immutable` to move them into `cacheControl` 🤔

jcready:
Now I’m curious, how is “immutable” useful for any cache duration? Could you point me to what you were reading?

dougwilson:
I was specifically reading through the Firefox bug for the implementation: https://bugzilla.mozilla.org/show_bug.cgi?id=1267474 and MDN https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control

Basically, it is a directive that tells the web browser that, if the browser has a valid cached version (which means it needs to have some directive to tell it to cache the response), it will just not send the revalidation requests until the resource expires out of the browser's cache.

jcready:
Interesting. I had (I guess incorrectly) assumed that “immutable” meant the resource could never change and that the maxAge part of it was only a fallback for clients that didn’t support the immutable part of the directive.

dougwilson:
Yea, I had the same assumption as well.

Issue #144, opened 2017-09-27
Regular Expression Denial of Service in mime 1.4.0


StanFisher:
According to NSP, there is a regular expression denial of service in one of your dependencies. The mime package needs to be updated to either version 1.4.1 or greater than or equal to 2.0.3.

This is rated as a high severity by Node Security Platform.

https://nodesecurity.io/advisories/535

dougwilson:
A fix will be released Monday Oct 6 (see statement at https://github.com/expressjs/express/issues/3431). The dependency may have an issue, but the issue does not impact this module. There is no patch available for the dependency without a semver-minor change, so we cannot simply release a patch version of this module.

dougwilson:
I would love to update it earlier as a patch if you know a method in which we can achieve a semver-patch upgrade of the dependency that has the fix.

dougwilson:
Right now `master` is staging a semver-minor release of this module. I'm talking about bumping from the current-published version of 1.3.4.

dougwilson:
Sorry, it's hard to describe anything now. I'm getting multiple emails a minute now about this before I had a change to do anything. Now my hands are basically being tied, unable to perform any code / releases / help and instead endlessly replying to emails.

PR #145, opened 2017-09-27
Bump mime to 1.4.1

dotLou:
This version fixes a security vulnerability flagged by snyk (https://github.com/expressjs/express/issues/3431) and nsp (https://nodesecurity.io/advisories/535) that's been fixed upstream.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "etag": "~1.8.1",
     "fresh": "0.5.2",
     "http-errors": "~1.6.2",
-    "mime": "1.4.0",
+    "mime": "1.4.1",
     "ms": "2.0.0",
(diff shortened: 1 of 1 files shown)


dougwilson:
It cannot be fixed right now. Duplicate of #144

dougwilson:
If you can make a pull request that can get us a semver-patch change, that would be amazing, though.

dougwilson:
To get a semver-patch, we need a release of mime that would fit between version 1.3.4 and 1.3.5, since mime has a semver-minor level change in their 1.3.x patch line (which is why we're not even on the latest 1.3.x). Right now `master` is preping the semver-minor release of this module, though, so I can actually merge this in now (at least it will keep more PRs from arriving).

PR #149, opened 2017-10-16
api: Added ability to set custom `fs` module

Marak:
This PR adds the ability to set a custom `fs` module for use inside of `send` module.

I needed this because I have another `fs` API compatible library for cloud files and I wanted to utilize `send` for serving static content from remote file sources. It's actually working great, so thank you!

Pseudo Example Usage:

```js
var send = require('send');
var remoteFs = require('the-remote-fs-module');
send.setFsModule(remoteFs);
// now use send() as normal, but will send remote files
```

Would you be willing to accept this PR into the library? It should be very harmless. Please let me know. I can make adjustments or add a simple test.

### DIFF
--- a/index.js
+++ b/index.js
@@ -69,6 +69,9 @@ var UP_PATH_REGEXP = /(?:^|[\\/])\.\.(?:[\\/]|$)/
 
 module.exports = send
 module.exports.mime = mime
+module.exports.setFsModule = function (vfs) {
+  fs = vfs;
+}
 
 /**
  * Return a `SendStream` for `req` and `path`.
--- a/index.js
+++ b/index.js
@@ -70,7 +70,7 @@ var UP_PATH_REGEXP = /(?:^|[\\/])\.\.(?:[\\/]|$)/
 module.exports = send
 module.exports.mime = mime
 module.exports.setFsModule = function (vfs) {
-  fs = vfs;
+  fs = vfs
 }
 
 /**


Marak:
@dougwilson - Could I get a yes or no on the **concept** of this PR? 

If yes, I will close and re-open with a cleaner PR.

Thanks!

dougwilson:
Sorry, I saw it come in, but it is in my backlog that I haven't been able to work on yet. I didn't really take much of a look at the PR yet, but the concept has been discussed several times before. I'm not against it, but the hardest thing is how this interacts with semver. For example, what exactly should the "fs" thing do (completely mimic all APIs and quirks for all Node.js versions from 0.8 to present)? If we start using another method on it, is that a semver major here? How do we set the right expectations for consumers to not break?

Marak:
No worries or rush. Good to hear it's a possibility for inclusion.

The PR itself is currently only three lines of code, which was enough to address my use-case.

I think the expectation is and always is that any custom `fs` module ( lets call it `vfs` ) passed in will always match the current version(s) of Node.js that `fs` and `send` module are being run / tested against. If there are any discrepancy or quirks in the `fs` API, it's up to the author of their `vfs` module to mimic that to support older versions.

I'd imagine in most cases emulating 90% of the `fs` API will be enough to get most `send` tests to pass. I'm on my way to getting 100% coverage on `send` with my custom `vfs` and modern versions of node.

**Short answer**: `send` library shouldn't care about semver of custom `fs` module, because it is outside scope of `send`. This PR is just about opening up API to allow injection of module.

dougwilson:
Right, I agree. More of how to manage the expectations of the user i.e. through docs or something.

Marak:
Understood. I'll make updates to this PR for: tests, documentation, and code style.

dougwilson:
As for this PR I took a look and does it make more sense for the vfs to be on a send instance rather than global? Global replacement would mean you have to account for all send usage within all your dependency tree, right?

Marak:
Eeek. Yes. I've only been running unit tests and testing in isolated child processes. 😨 

Instance based is definitely the way to go. That would make the API better with something like `options.fs` to signify a custom `fs` module.

Marak:
Closing PR. Will open a new PR with instance base API, documentation, and tests.

PR #151, opened 2017-12-01
Update mime

castilloandres:
Update mime package from `v1.4.1` to `v1.6.0` to fix -> https://github.com/expressjs/express/issues/3486

PR #154, opened 2018-01-05
Update mime

devongovett:
#151 but with corrected history

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -7,6 +7,19 @@ unreleased
     - deps: depd@~1.1.2
     - deps: setprototypeof@1.1.0
     - deps: statuses@'>= 1.5.0 < 2'
+  * deps: mime@1.6.0
+    - Add extensions for JPEG-2000 images
+    - Add new `font/*` types from IANA
+    - Add WASM mapping
+    - Update `.bdoc` to `application/bdoc`
+    - Update `.bmp` to `image/bmp`
+    - Update `.m4a` to `audio/mp4`
+    - Update `.rtf` to `application/rtf`
+    - Update `.wav` to `audio/wav`
+    - Update `.xml` to `application/xml`
+    - Update generic extensions to `application/octet-stream`:
+      `.deb`, `.dll`, `.dmg`, `.exe`, `.iso`, `.msi`
+    - Use mime-score module to resolve extension conflicts
   * deps: ms@2.1.1
     - Add `week`/`w` support
     - Fix negative number handling
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "etag": "~1.8.1",
     "fresh": "0.5.2",
     "http-errors": "~1.7.1",
-    "mime": "1.4.1",
+    "mime": "1.6.0",
     "ms": "2.1.1",
     "on-finished": "~2.3.0",
     "range-parser": "~1.2.0",


dougwilson:
Thanks! We'll discuss this in the coming week to see when a new minor version will get cut 👍

dougwilson:
We should first evaluate the impact of the changes to the mime types this introduces (like our usual assessment to the mime types). Here is the diff (only changed ones):
```diff
-bdoc   application/x-bdoc
+bdoc   application/bdoc
-bmp    image/x-ms-bmp
+bmp    image/bmp
-deb    application/x-debian-package
+deb    application/octet-stream
-dll    application/x-msdownload
-dmg    application/x-apple-diskimage
+dll    application/octet-stream
+dmg    application/octet-stream
-exe    application/x-msdownload
+exe    application/octet-stream
-iso    application/x-iso9660-image
+iso    application/octet-stream
+jp2    image/jp2
+jpf    image/jpx
+jpg2   image/jp2
-jpm    video/jpm
+jpm    image/jpm
+jpx    image/jpx
-m4a    audio/x-m4a
+m4a    audio/mp4
-msi    application/x-msdownload
+msi    application/octet-stream
-org    text/x-org
+org    application/vnd.lotus-organizer
-pdb    application/x-pilot
+pdb    application/vnd.palm
-prc    application/x-pilot
+prc    application/x-mobipocket-ebook
-ra     audio/x-realaudio
+ra     audio/x-pn-realaudio
+raml   application/raml+yaml
-rtf    text/rtf
+rtf    application/rtf
-ttc    application/x-font-ttf
-ttf    application/x-font-ttf
+ttc    font/collection
+ttf    font/ttf
-wav    audio/x-wav
+wasm   application/wasm
+wav    audio/wav
-wmz    application/x-msmetafile
-woff   application/font-woff
-woff2  application/font-woff2
+wmz    application/x-ms-wmz
+woff   font/woff
+woff2  font/woff2
-xml    text/xml
+xml    application/xml
```

Looks like there is a high level of impact potentially here to existing servers, more than I have ever seen in a `mime` upgrade. Things like `exe`, `iso`, `msi`, `xml`, `rtf`, `wav`, `bmp` looks like the largest changes.

castilloandres:
Closed #151(https://github.com/pillarjs/send/pull/151) will be following up on this PR 👍 @dougwilson  @devongovett

dougwilson:
So new Express.js minor is looking at Feb release. Ideally this will be included in that release if the mime changes are found to not be impactful to existing servers. If any are, we have time to make changes prior to the minor, so the early it's determined, the better 👍

robjtede:
+1 for this.
WASM support in static servers like `browser-sync` and `lite-server` that depend on this down the tree would be amazing.

dougwilson:
If the change was simply just adding WASM it would have been released already. Instead, it is changing a lot of widely used extensions too. If anyone wants to see it land sooner, please help assess the impact of the changes listed above :+1:

robjtede:
Yeah, makes since given this package is so fundamental.

dougwilson:
So I realized that we took on some changes with mime 1.4.1 because of a security issue, so had to accept the churn in that version. There have been some issues opened in the main Express repo with issues due to that churn, but we didn't want to revert due to security. What I'm getting at is a new comparison between 1.3.4 and 1.6.0 instead of 1.4.1 and 1.6.0. Here is the new comparison of changes types:
```patch
-deb	application/x-debian-package
+deb	application/octet-stream
-dll	application/x-msdownload
+dll	application/octet-stream
-dmg	application/x-apple-diskimage
+dmg	application/octet-stream
-exe	application/x-msdownload
+exe	application/octet-stream
-iso	application/x-iso9660-image
+iso	application/octet-stream
-jpm	video/jpm
+jpm	image/jpm
-markdown	text/x-markdown
+markdown	text/markdown
-md	text/x-markdown
+md	text/markdown
-msi	application/x-msdownload
+msi	application/octet-stream
-otf	font/opentype
+otf	font/otf
-ttc	application/x-font-ttf
+ttc	font/collection
-ttf	application/x-font-ttf
+ttf	font/ttf
-wav	audio/x-wav
+wav	audio/wav
-wmz	application/x-msmetafile
+wmz	application/x-ms-wmz
-woff	application/font-woff
+woff	font/woff
-woff2	application/font-woff2
+woff2	font/woff2
```

All the font changes are definitely known and may or may not cause issues, but I think they are acceptable. Removing those leaves us with the following:
```patch
-deb	application/x-debian-package
+deb	application/octet-stream
-dll	application/x-msdownload
+dll	application/octet-stream
-dmg	application/x-apple-diskimage
+dmg	application/octet-stream
-exe	application/x-msdownload
+exe	application/octet-stream
-iso	application/x-iso9660-image
+iso	application/octet-stream
-jpm	video/jpm
+jpm	image/jpm
-markdown	text/x-markdown
+markdown	text/markdown
-md	text/x-markdown
+md	text/markdown
-msi	application/x-msdownload
+msi	application/octet-stream
-wav	audio/x-wav
+wav	audio/wav
-wmz	application/x-msmetafile
+wmz	application/x-ms-wmz
```

Further, the two markdown I think are a normal move forward, so removing:
```patch
-deb	application/x-debian-package
+deb	application/octet-stream
-dll	application/x-msdownload
+dll	application/octet-stream
-dmg	application/x-apple-diskimage
+dmg	application/octet-stream
-exe	application/x-msdownload
+exe	application/octet-stream
-iso	application/x-iso9660-image
+iso	application/octet-stream
-jpm	video/jpm
+jpm	image/jpm
-msi	application/x-msdownload
+msi	application/octet-stream
-wav	audio/x-wav
+wav	audio/wav
-wmz	application/x-msmetafile
+wmz	application/x-ms-wmz
```

So we can split those into two groups: ones that are now `application/octet-stream`: `deb`, `dll`, `dmg`, `exe`, `iso`, `msi`; and ones that changed in general: `jpm`, `wav`, and `wmz`.

My understanding is that `jpm` wasn't correct in the first place, so that sound then be OK.

Looking at `wmz` it seems like that extension has meant so many various things over time, so probably whatever it is coming through now is meaningless anyway.

The `wav` change seems big, though it's just removing the `x-` classification from the type. Research with browsers seems to be that `audio/wav` is the better choice to serve under vs `audio/x-wav` so makes sense.

Now that just leaves all the ones that became `application/octet-stream`. Just looking around there are no "real" types for any of those in the first place. If this will actually be an issue remains to be seen, since most clients (like web browsers) won't even handle any of those types of files, and simply download + save, as which point the served type is lost. So it does seem fine.

TL;DR through review of the changes looks acceptable so it will absolutely be included in the next minor release 👍

ColinEberhardt:
Any chance this could be merged / released? WebAssembly, which uses the `wasm` files extension and should be server with the mime type `application/wasm`. If the mime type is incorrect, streaming compilation fails within the browser.

rotemdan:
Latest version of `send` (at least as seen through `live-server`), serves `.wasm` files with type `octet-stream` instead of `application/wasm`, [as required in the standard for streaming WASM modules](https://webassembly.github.io/spec/web-api/index.html#streaming-modules)). Not having the right MIME type causes an error (by design):

> If the Response is not CORS-same-origin, does not represent an ok status, or does not match the `application/wasm` MIME type, the returned promise will be rejected with a TypeError.

Is this PR planned to fix this? Will this be eventually merged?

dougwilson:
I (or someone who wants to volunteer) just needs to rebase this PR so it is mergable to merge. As for releasing send releases are tied to Express releases. A new Express minor is not yet being prepped since the research on this issue was completed for the mime change impacts, so I haven't rebased the PR myself yet since I don't have a timeline for a new Express minor version this would go into.

dougwilson:
This was noted in some other issues but not in this PR, so I want to add it here if people are find their way here:

The types provided by this module are the _default_ types, not the only possible types. This module exposes the underlying Mime instance you can configure with your own types to add / remove / change them as-needed. This module updates to the latest mime ever minor release and web standards move and the mime module didn't provide the wasm type map at the time of our last minor.

To add this (or any other type you want / need), just add a single line your program:

    send.mime.types['wasm'] = 'application/wasm'

rotemdan:
@dougwilson 

Thanks for the quick response,

I'm not involved in the development of `live-server`, but digging through the code I figured out it was using `send` to resolve mime types, so eventually I tried to see if I can get to the "root" cause. 

`live-server` doesn't have a mechanism for mime type exceptions (yet, it is a [requested feature](https://github.com/tapio/live-server/issues/230) though) and doesn't seem to make any exceptions of its own internally, and even if it did, I would in turn need to ask to the developers of the [`vscode-live-server`](https://github.com/ritwickdey/vscode-live-server) extension (which doesn't have a mechanism for MIME type exceptions, and I'm not involved in, as well) to apply this special case. And they in turn possibly point back and say that either `live-server` or `send` should have had it in the first place etc. 

As you can see there's a complex chain of dependencies here that's not very easy to control.

dougwilson:
> As you can see there's a complex chain of dependencies here that's not very easy to control.

I definitely understand. We already have this PR and are planning to add it. As much as is my control I try to give the users of my module all the tools they need to kot wait for me. Just commenting here creates unnecessary noise when a large part of the issue is with those other modules preventing you from using the features this module offers you. If those are unwilling to allow you to customize the mime types, there is nothing further I can do to help you. You'll either need to use something else or wait until the next minor Express release planning begins and this gets included in that process.

robjtede:
May I ask why a release of express is a block on this PR?

LucasVos:
I was looking for wasm mime type support in `live-server` and end-up here. Any news or schedule about this PR?

PR #155, opened 2018-01-17
Upgrade to mime@1.5.0.

ndarilek:
There are newer versions of this package available, but 1.5.0 includes support for the application/wasm mime type, which I need in a dependency several packages removed. If you wanted to reject this in favor of a major version bump then I'd understand, but this seems to be the smallest change that would meet my requirements. Thanks!

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "etag": "~1.8.1",
     "fresh": "0.5.2",
     "http-errors": "~1.6.2",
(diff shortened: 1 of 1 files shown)


dougwilson:
Duplicate of #154 if you want to help evaluate the impact there 👍 A new Express.js minor is scheduled for Feb, so ideally it would be included then if it's not founds to be impactful based on all those type changes.

Issue #126, opened 2017-01-10
Why remove Content-Security-Policy header?


felipeochoa:
Is there a reason for removing the CSP header? I'm using this module to serve the script for a `WebWorker` and need to keep the CSP header on there to allow the worker to make XHR calls.

This header is removed in [removeContentHeaderFields](https://github.com/pillarjs/send/blob/master/index.js#L328). It looks like a farily generic check, so I'm not even sure if you're removing this header on purpose.

dougwilson:
The module is trying to follow what is specified for generating a 304 https://tools.ietf.org/html/rfc7232#section-4.1

dougwilson:
We should dig around through the RFCs to see what should happen w.r.t that header and a 304.

felipeochoa:
Yeah I’m not sure whether CSP assists in “guiding cache invalidation.” This is outside my area of expertise, but this is the closest I was able to find: https://github.com/w3c/webappsec-csp/issues/161 indicates sending CSP headers on 304 is valid.

dougwilson:
Thanks, that helped a lot! From reading that, it sounds like there is nothing to change here. Even the fetch client specifies what this module is doing is correct. You can see when reading through the specs linked to from the lost recent comment that the steps in stage 21 is (1) load the cached response (2) if cannot load, network error (3) take any header from the network response and update the cached header.

This means that, just like it says in the http rfc, not including the header is not incorrect and simply means the browser will see the header value as it was I the 200 response it loaded up from it's cache response.

Both the http spec and that fetch spec are in agreement with as implemented here with the added statement that the http spec calls out that the server SHOULD not send any extra headers if possible, which is what we're doing.

dougwilson:
Sorry, didn't mean to close, this is still in discussion.

dougwilson:
So additional reading it looks like this module is behaving perfectly fine according to the specs and the only real issue in web browsers today with CSP + 304 is that if you're using a nonce in the CSP, if you send the CSP header with the 304 in some browsers, everything will break, so because of that bug, you actually _do not want to send the CSP header with the 304_, this is because you'll send a new nonce, yet because the content was loaded from the browser cache, the nonce will no longer match.

PR #164, opened 2018-08-07
[URGENT][SECURITY] Keep Content-Security-Policy http header in 304 requests

pavelkornev:
Hi there,

we have found out that Content-Security-Policy http-headers are being dumped in case of 304 (Not Modified) requests which brings security problems/risks for the enterprise applications and makes it impossible to use this this library in a production environment.

Please consider to keep these very important http headers in any kind of requests.

Thanks,
Pavel

### DIFF
--- a/index.js
+++ b/index.js
@@ -360,7 +360,7 @@ SendStream.prototype.removeContentHeaderFields = function removeContentHeaderFie
 
   for (var i = 0; i < headers.length; i++) {
     var header = headers[i]
-    if (header.substr(0, 8) === 'content-' && header !== 'content-location') {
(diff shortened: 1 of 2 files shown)


pavelkornev:
I've the closed thread - https://github.com/pillarjs/send/issues/126

It's basically the same security story as discussed here with CORS - https://github.com/web-platform-tests/wpt/pull/1400 (linked to https://github.com/w3c/webappsec-csp/issues/161)

And as we see, browsers do not use old CSP http headers from the cache. So, yes, with the spec it might  not be clear what to do at this moment, but as a result by removing CSP headers we put data sensitive applications at a serious risk as it would be when dropping CORS.

@dougwilson do you have any objections?

Issue #160, opened 2018-06-08
Allow using alternate sources of files


LynnKirby:
It would be useful to abstract the filesystem API so that, for example, in-memory files can be served.

I have a proof-of-concept (or even solution?) at 6c87c364d33e0c52c996e08576edd15913a33fae. It adds an `fs` property to the options which requires `fs.stat` and `fs.createReadStream` functions. Set the environment variable `USE_MEMFS` before running tests and the test will use [memfs](https://github.com/streamich/memfs). [memory-fs](https://github.com/webpack/memory-fs) doesn't work because the `fs.Stats` objects it creates only has functions attached and not all of the usual properties.

Related: expressjs/serve-static#83

dougwilson:
That PoC seems to effectively be a duplicate of #149 and would have the same concerns. The OP of #149 never followed back up on the asks. Would you be able to incorporate them into your implementation?

dougwilson:
> memory-fs doesn't work because the fs.Stats objects it creates only has functions attached and not all of the usual properties.

This sounds like my main concern with just accepting `fs` as an option: how do users know what works and what does not and how can we prevent accepting a `fs` option that will suddenly blow up later in the implementation after the initial construction?

LynnKirby:
Documentation is easy, the hard part is deciding on the requirements.

The concern about having to increase the major version if more of `fs` is used is valid. I can think of two solutions:

1. Mark the feature as experimental and therefore may change between minor versions.

2. Just accept the possibility of major version changes. I looked at the library history and even in 0.0.2 only those two functions were being used. There hasn't been any change for six years (maybe, I didn't check everywhere), so is it reasonable to think this problem won't happen?

dougwilson:
So we are intending to implement #122 to resolve a race condition which is definitively a planned change to the fs API usage currently. I think there is another open issue to use another new part of the fs API as well.

When a new major version gets cut, that becomes a larger burden because the Express.js project exposes this module as a core piece. That would that Express.js would also have to bump the major to use it and of course would mean maintaining the prior major version of this module for the LTS lifetime of the corresponding Express (which is that the prior major version of Express is supported for 1 year after a new major is released). Obviously this will happen, but it would be awesome if it doesn't require the frequency of majors to be more than one per year (which is how Node.js itself functions).

As for marking as experimental, I'm not quite sure I'm following there. Should we match how Node.js does it and make it inaccessible unless node is started with a specific command line flag?

LynnKirby:
[...]

PR #174, opened 2018-10-23
Proposal: Add host option to override node fs and path modules

barak007:
Hi,
I have a use-case that requires me to serve files from memory in my tests.
Currently, I use `express.static()` from `serve-static` which uses this package. Now, I want to keep the same API but enable passing an option to override the default host file-system and path resolution.
There is an [issue](expressjs/serve-static#83) in `express-static` that requests the same thing but hasn't seen much activity.

About the tests missing from this pull request.
Since there is no changed behavior, all of the current tests are passing. What I would really want to do is wrap all the existing test-cases in a factory and re-run them using an external host.
Would you have any objections to this test plan? Would love to get this in.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,12 @@ not the actual file-system path).
 
 #### Options
 
+##### host
+
+Enable override of the default nodejs `fs` and `path`.
+use this option to enable in memory or remote sources support 
+Default value: `{ fs: require('fs'), path: require('path') }`
+
 ##### acceptRanges
 
 Enable or disable accepting ranged requests, defaults to true.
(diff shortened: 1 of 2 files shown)


dougwilson:
Sorry, this is not the correct implementation for all the reasons from the previously closed PRs and issues on this.

Please move the discussion/proposal to https://github.com/pillarjs/send/issues/160 . If someone can provide a proposal that addresses these, that would be the way to move forward.

I'm actually working on a security fix currently that would complete rewrite which fs methods are called, even apart from the changes PR #125 makes. This PR would make those types of changes no longer backwards compatible as our exact usage of the internal fs becomes public as folks will end up only passing in the bare minimum to make their use case work.

See past discussion #149 as well.

PR #177, opened 2019-01-04
case sensitive file handling

sebbi08:
This is ment to solve this issue https://github.com/expressjs/serve-static/issues/93.
It contains a fix to handel files on in case insensitive file system case sensitive.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,6 +119,15 @@ Serve files relative to `path`.
 Byte offset at which the stream starts, defaults to 0. The start is inclusive,
 meaning `start: 2` will include the 3rd byte in the stream.
 
+##### caseSensitive
+
+When set to `true` all the requests against the file system are made case sensitive even on case insensitive file systems.
+This effects not only the file name but the whole path.
+When set to `false` the default behavior of the filesystem is used.
+
+The default value is `false`
+
+
 #### Events
 
 The `SendStream` is an event emitter and will emit the following events:
--- a/index.js
+++ b/index.js
@@ -729,7 +729,7 @@ SendStream.prototype.sendFile = function sendFile (path) {
     }
     if (err) return self.onStatError(err)
     if (stat.isDirectory()) return self.redirect(path)
-    if (self.options.caseSensitive && !fileExistsWithCaseSync(path)) {
+    if (self._caseSensitive && !fileExistsWithCaseSync(path)) {
       return self.error(404)
     }
     self.emit('file', path, stat)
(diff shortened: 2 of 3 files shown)


sebbi08:
is there any diference betwene the tracis ci coverage reporter and the local one localy i get the following coverage results : 

Statements   : 99.77% ( 425/426 )
Branches     : 96.36% ( 238/247 )
Functions    : 100% ( 49/49 )
Lines        : 100% ( 416/416 )

index.js
```diff
@@ -1127,3 +1137,20 @@ function setHeaders (res, headers) {
...
+ * Check if the given file incusive the path exits with the check on case sensitivity
+ * @param filepath the file to be checked
+ * @return {boolean} true when a file with the same case exists false if not
+ */
+function fileExistsWithCaseSync (filepath) {
+  var dir = path.dirname(filepath)
+  if (dir === path.dirname(dir)) {
+    return true
+  }
+  var filenames = fs.readdirSync(dir)
```
dougwilson, line 1151: We should really do this using an async method; sync methods will block the event loop and processing of all I/O in the process while the read occurs, and it could be against a network share, for example.

README.md
```diff
@@ -119,6 +119,15 @@ Serve files relative to `path`.
 Byte offset at which the stream starts, defaults to 0. The start is inclusive,
 meaning `start: 2` will include the 3rd byte in the stream.
 
+##### caseSensitive
+
+When set to `true` all the requests against the file system are made case sensitive even on case insensitive file systems.
+This effects not only the file name but the whole path.
```
dougwilson, line 125: (this is just a comment) I assume by "whole path" that includes what was provided also as the `root` option? If so, it should probably call that out, as there is the input named `path`, which may be mistake for "the whole path" in this wording.

README.md
```diff
@@ -119,6 +119,15 @@ Serve files relative to `path`.
...
+When set to `false` the default behavior of the filesystem is used.
```
dougwilson, line 126: Passing in `caseSensitive: false` does have a weird feel to the API... it feels like it should thus *not* be case-sensitive, vs depend on the file system. I'm trying to think of a good API name for this functionality, and don't have a good one.

Would it be possible to maybe even make `caseSensitive: false` work? For example, to have this option be `true`, or `false, or `'fs'`? If not, as I cannot think of a good API name that would make sense as true/false, maybe use a string like `'fs'` instead of accepting `false`?

index.js
```diff
@@ -546,6 +548,7 @@ SendStream.prototype.pipe = function pipe (res) {
 
     // join / normalize from optional root dir
     path = normalize(join(root, path))
+    root = normalize(root + sep)
```
dougwilson, line 551: What is this doing, exactly?

PR #184, opened 2019-05-11
Deprecated res._headers method replaced with Node12 complaint res.getHeaders

Doc999tor:
res._headers causes `[DEP0066] DeprecationWarning: OutgoingMessage.prototype._headers is deprecated in Node.js 12 runtime warning`
Replaced with public `res.getHeaders` method
Links: [Node docs](https://nodejs.org/api/deprecations.html#deprecations_dep0066_outgoingmessage_headers_outgoingmessage_headernames), [res.hetHeaders](https://nodejs.org/api/http.html#http_response_getheaders)

### DIFF
--- a/index.js
+++ b/index.js
@@ -1002,7 +1002,7 @@ function decode (path) {
 
 function getHeaderNames (res) {
   return typeof res.getHeaderNames !== 'function'
-    ? Object.keys(res._headers || {})
+    ? Object.keys(res.getHeaders || {})
     : res.getHeaderNames()
 }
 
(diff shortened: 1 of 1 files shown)


dougwilson:
So I just ran the rest suite in Node.js 12 and got no deprecation warning with the existing code. This is because res.getHeaderNames exists in Node.js 12 so the branch you are changing is never executed.

I can reopen if you can provide the steps to reproduce the deprecation message and show that it is cause by this module.

PR #191, opened 2020-02-23
deps: http-errors@~1.7.3

UlisesGascon:
Please review it @dougwilson 😄 

This is part of the Express deps upgrade for 4.x. [See](https://github.com/expressjs/express/issues/4171#issuecomment-590061349)


HISTORY.md
```diff
@@ -1,3 +1,8 @@
+unreleased
+===================
+
+  * deps: http-errors@~1.7.3
```
dougwilson: We should add a sub bullet here with the inherits change that is part of this bump since it will be a visible change when installing send.

HISTORY.md
```diff
@@ -1,3 +1,8 @@
+unreleased
+===================
```
dougwilson: You'll want to reduce the number of `=` to match the length of the header.

UlisesGascon: Great catch! I fixed at d34bc21

HISTORY.md
UlisesGascon: Makes sense! I fixed at d34bc21

UlisesGascon:
I will update `send` version for `serve-static` in a separate PR once this change is released 👍

dougwilson:
Apologies, this PR got lost in the shuffle and new dep updater updated it before I realized there was a PR open.

PR #192, opened 2020-03-05
Switch to mime-types from mime

LinusU:
This was picked up from the Express 5.0 todo-list: expressjs/express#2237

> Use `mime-types` instead of `mime`

`accepts` is already using `mime-type` so currently both are being pulled in to Express:

```bash
$ npm ls mime-types
express@5.0.0-alpha.7 /Users/linus/coding/express
├─┬ accepts@1.3.7
│ └── mime-types@2.1.26 
├─┬ supertest@3.3.0
│ └─┬ superagent@3.8.3
│   └─┬ form-data@2.5.1
│     └── mime-types@2.1.26  deduped
└─┬ type-is@1.6.18
  └── mime-types@2.1.26  deduped

$ npm ls mime      
express@5.0.0-alpha.7 /Users/linus/coding/express
├─┬ send@0.16.2
│ └── mime@1.4.1 
└─┬ supertest@3.3.0
  └─┬ superagent@3.8.3
    └── mime@1.4.1  deduped
```

Note that `supertest` is a dev-dependency so after this change only `mime-types` should be present when adding express as a dependency.

-----

Proposal for changelog:

```markdown
0.18.0 / 2020-03-XX
===================

  * **Breaking:** Switch to `mime-types` from `mime`

    Migration guide:

    The mime type database is no longer extendable at runtime, and thus the `.mime` export have been removed.
```

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -7,6 +7,7 @@
   * Remove `send.index()` -- use `index` in `options`
   * Remove `send.maxage()` -- use `maxAge` in `options`
   * Remove `send.root()` -- use `root` in `options`
+  * Use `mime-types` for file to content type mapping -- removed `send.mime`
 
 0.17.2 / 2021-12-11
 ===================
--- a/README.md
+++ b/README.md
@@ -133,15 +133,6 @@ The `SendStream` is an event emitter and will emit the following events:
 The `pipe` method is used to pipe the response into the Node.js HTTP response
 object, typically `send(req, path, options).pipe(res)`.
 
-### .mime
-
-The `mime` export is the global instance of of the
-[`mime` npm module](https://www.npmjs.com/package/mime).
-
-This is used to configure the MIME types that are associated with file extensions
-as well as other options for how to resolve the MIME type of a file (like the
-default type to use for an unknown file extension).
-
 ## Error-handling
 
 By default when no `error` listeners are present an automatic response will be
@@ -210,20 +201,22 @@ server.listen(3000)
 ### Custom file types
 
 ```js
+var extname = require('path').extname
 var http = require('http')
 var parseUrl = require('parseurl')
 var send = require('send')
 
-// Default unknown types to text/plain
-send.mime.default_type = 'text/plain'
-
-// Add a custom type
-send.mime.define({
-  'application/x-my-type': ['x-mt', 'x-mtt']
-})
-
 var server = http.createServer(function onRequest (req, res) {
   send(req, parseUrl(req).pathname, { root: '/www/public' })
+    .on('headers', function (res, path) {
+      switch (extname(path)) {
+        case '.x-mt':
+        case '.x-mtt':
+          // custom type for these extensions
+          res.setHeader('Content-Type', 'application/x-my-type')
+          break
+      }
+    })
     .pipe(res)
 })
 
--- a/index.js
+++ b/index.js
@@ -21,7 +21,7 @@ var escapeHtml = require('escape-html')
 var etag = require('etag')
 var fresh = require('fresh')
 var fs = require('fs')
-var mime = require('mime')
+var mime = require('mime-types')
 var ms = require('ms')
 var onFinished = require('on-finished')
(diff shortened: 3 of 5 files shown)


LinusU:
> Since this is a breaking change can this target the 1.x branch along with the other breaking changes for 1.x?

Sure thing ✔️  

> Also, we need to at least document some method for users to change the mime types for files based on extension... that would be a very large and strange API gap to introduce, as setting that type of thing is common among all static file servers.

Sounds good! I'm guessing that we should introduce an API specifically for this? I'll post a suggestion on it soon ✔️  

> Please include the updates to the history file as part of the PR itself :)

Will do ✔️

dougwilson:
> Sounds good! I'm guessing that we should introduce an API specifically for this? I'll post a suggestion on it soon ✔️

We may need to, yes. Currently mime-types itself doesn't have one, but we probably need to do _something_ as it is quite the feature gap. The mime module was archived for a while but has been since unarchived. The motivation behind this change is partly due to that and to simplify the dep tree.

I think we may also decide there is too much work to do still to throw it into express 5 as well, as I think we all agree a top goal is to prevent further slipping of the release, lol.

Sorry I don't have a direct opinion right now; just quickly typing at work and will give more thought, especially if you have more to post before I'm out tonight.

LinusU:
> I think we may also decide there is too much work to do still to throw it into express 5 as well, as I think we all agree a top goal is to prevent further slipping of the release, lol.

100%, I don't want this to hold up the release at all 👍 

> Sorry I don't have a direct opinion right now

No worries!

-----

Looking at the current API, I'm not a fan of extending global variables. Especially since any library can import `mime` and add something to it, but that might not always be the same `mime` that `send` will use depending on how the dependency tree looks. I think it lends itself to produce bugs that are hard to catch...

One idea for an API would be something like this:

- In `send` we add an option to pass in a mime-type database/source object. In a breaking change, we could default this to `require('mime-db')`.
- In `serve-static` we add the same option, which would just be passed to `send`. In a breaking change, we could bump the `send` version so that it defaults to `mime-db`.
- In Express we add the same option as a global `.set()` option that, if set, gets passed to `serve-static`/`send`. In a breaking change, we could bump the `serve-static`/`send` version so that it defaults to `mime-db`.

On top of this we would probably also need an `extend` function that can extend an existing mime database object, so that people can easily add there own mime types.

The usage would look something like this:

**Send**

```js
const send = require('send')
const mimeDB = require('mime-db')

const mimeTypes = { ...mimeDB, 'application/linusu': { ext: ['.foo'] } }

// ...

send(req, 'linusu.foo', { mimeTypes })
```

**serve-static**

```js
const serveStatic = require('serve-static')
const mimeDB = require('mime-db')

const mimeTypes = { ...mimeDB, 'application/linusu': { ext: ['.foo'] } }
const serve = serveStatic('public/linusu', { mimeTypes })

// ...

serve(req, res, (err) => console.error(err.stack))
```

**express**

```js
const express = require('express')
const mimeDB = require('mime-db')

const app = express()
const mimeTypes = { ...mimeDB, 'application/linusu': { ext: ['.foo'] } }

app.set('mime types', mimeTypes)

app.get('/', (req, res) => {
  res.sendFile('linusu.foo')
})

app.use(express.static('public/linusu'))

// ...
```

-----

The beauty with this approach:

- We don't _have_ to make it a breaking change
  - (the only breaking change would be defaulting to the new mime-db)
- No global variables, but still easy to configure your entire Express project
- Users can use alternative mime databases
  - (e.g. one with only the types that they are going to use)

What do you think a good approach to move this forward is? Open an issue on expressjs/discussions to start a discussion on the API design? or submit draft PRs implementing my suggestion? or something else? ☺️

dougwilson (approved):
I know this issue is old, and kind of fell off the radar. I'm going to last this as-is on the 1.0 release branch here, but you make a lot of good points in your last issue. I figure we could probably get that implemented prior to 1.0, or even after, it doesn't really matter, since it should be semver minor.

Issue #200, opened 2020-11-24
lookup not a function


WilliamRagstad:
I get the error:
```
C:\Users\willi\odrive\Google Drive\- KTH\- Kurser\Termin 2\DH2642 Interaktionsprogrammering och dynamiska webben\Projekt\DH2642-Project\node_modules\send\index.js:840
  var type = mime.lookup(path)
                  ^

TypeError: mime.lookup is not a function
    at SendStream.type (C:\Users\willi\odrive\Google Drive\- KTH\- Kurser\Termin 2\DH2642 Interaktionsprogrammering och dynamiska webben\Projekt\DH2642-Project\node_modules\send\index.js:840:19)
    at SendStream.send (C:\Users\willi\odrive\Google Drive\- KTH\- Kurser\Termin 2\DH2642 Interaktionsprogrammering och dynamiska webben\Projekt\DH2642-Project\node_modules\send\index.js:627:8) 
    at onstat (C:\Users\willi\odrive\Google Drive\- KTH\- Kurser\Termin 2\DH2642 Interaktionsprogrammering och dynamiska webben\Projekt\DH2642-Project\node_modules\send\index.js:729:10)
    at FSReqCallback.oncomplete (node:fs:192:5)
```

On this location
https://github.com/pillarjs/send/blob/de073ed3237ade9ff71c61673a34474b30e5d45b/index.js#L838

Please change to:
```
var type;
if (mime.lookup) type = mime.lookup(path)
```

dougwilson:
Hi @WilliamRagstad you likely have a `node_modules` folder that is not correct. You can delete it and run `npm install` again to fix it, and run `npm ls` to look for any errors about bad dependencies.

To serve static files the type needs to be there, and if you don't want to use the automatic type lookup, if you set the `content-type` header yourself, this module will not attempt to look up the type.

PR #201, opened 2020-12-01
Fix Issue #200

WilliamRagstad:
**Error:** `mime.lookup is not a function`
**Solution:** Function is checked before use. No difference in functionality but fixes critical error for some dependent modules.

I do not know why `mime.lookup` is undefined, but this has stopped me from using other packages that are dependent on send and has required me to continuously change this piece of code each time to be able to build my project. I would be grateful if this was merged and included in the next release.

### DIFF
--- a/index.js
+++ b/index.js
@@ -835,7 +835,8 @@ SendStream.prototype.type = function type (path) {
 
   if (res.getHeader('Content-Type')) return
 
-  var type = mime.lookup(path)
+  var type;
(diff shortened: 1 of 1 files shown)


dougwilson:
This just gets around one error you will see.this module and modules that use it expect the mime function to work and the content-type header needs to be in your responses for your assets to load properly in web browsers.

If you can include a change to the test suite that validates this change we can reconsider.

PR #213, opened 2022-07-03
support for /proc/* files

mleblebici:
/proc/* files could not be read, because fs.stat was returning 0 size. With new getSize function, their size information is correct, so they can now be read.

### DIFF
--- a/index.js
+++ b/index.js
@@ -721,6 +721,9 @@ SendStream.prototype.sendFile = function sendFile (path) {
     }
     if (err) return self.onStatError(err)
     if (stat.isDirectory()) return self.redirect(path)
+    if (stat.size === 0) {
+      stat.size = getSize(path);
+    }
     self.emit('file', path, stat)
     self.send(path, stat)
   })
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1141,3 +1144,38 @@ function setHeaders (res, headers) {
...
+			if (length <= tmpBufMaxLen) {
+				tmpBuf = newBuf;
+			}
+
+			buf.copy(newBuf);
+			buf = newBuf;
+		}
+	} while (bytesRead !== 0);
+	fs.closeSync(fd);
+	return buf.toString('utf8', 0, pos).length;
```
dougwilson, line 1179: This will return the wrong size if the data contains any utf8 charachers, as the length is the number of characters in this line, but what is needed is the number of bytes.

index.js
```diff
@@ -1141,3 +1144,38 @@ function setHeaders (res, headers) {
...
+	const fd = fs.openSync(path, 'r');
+	const tmpBufMinLen = 4096 * 2;
+	const tmpBufMaxLen = 4096 * 8;
+	let tmpBuf = Buffer.allocUnsafeSlow(tmpBufMinLen);
+	let pos = 0;
+	let bytesRead;
+	let buf = tmpBuf;
+	let length = buf.length;
+	do {
+		bytesRead = fs.readSync(fd, buf, pos, buf.length - pos, null);
```
dougwilson, line 1164: Performing sync operations will halt processing of parallel requests of the http server. You will need to use the async version of the functions, not the sync versions.

dougwilson (changes requested):
A general comment is that it seems when stat reads the file size as zero, this new code will read the entire contents of the file in to memory. If that is necessary, there should be some kind of upper limit on it so it does not cause the web server to oom on a file. You probably can just track the read chunk lengths without sorting a complete buffere to get the length of.

index.js
```diff
@@ -1012,3 +1144,38 @@ function setHeaders (res, headers) {
...
+			if (length <= tmpBufMaxLen) {
+				tmpBuf = newBuf;
+			}
+
+			buf.copy(newBuf);
+			buf = newBuf;
+		}
+	} while (bytesRead !== 0);
+	fs.closeSync(fd);
+	return buf.toString('utf8', 0, pos).length;
```
aelmardhi, line 1179: Shouldn't pos have the actual size

PR #214, opened 2022-09-08
Support must-revalidate

khoazero123:


wesleytodd:
@khoazero123 are you interested in writing some docs and tests for this? If not I might just close this until someone can pick up the remaining work which would be required to land it.

PR #224, opened 2024-01-20
Do not serve files when path ends with /

rmhaiderali:
When `index` option is set to `false` and the environment is Windows, requests that end with a `/` might result in serving files. This happens because in Windows, when accessing a file or directory, if the path ends with `/`, the system first checks for a directory. If no directory is found, it then checks if a file with the same name exists, ignoring the `/` at the end of the path. This enables access to a file unexpectedly, as if the path did not conclude with `/`.

The proposed fix addresses this issue by introducing a check to ensure that the path does not ends with `/` before serving a file. For example, in the given case, file is being accessed when it should not be, and the proposed change aims to rectify this behavior.

`index.js`
```JS
var http = require("http");
var send = require("send");

var server = http.createServer(function onRequest(req, res) {
  send(req, "./index.html/", { index: false }).pipe(res);
});

server.listen(3000);
```
`index.html`
```HTML
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="UTF-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
    <title>Document</title>
  </head>
  <body>
    HTML File
  </body>
</html>
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -605,12 +605,16 @@ SendStream.prototype.sendFile = function sendFile (path) {
 
   debug('stat "%s"', path)
   fs.stat(path, function onstat (err, stat) {
-    if (err && err.code === 'ENOENT' && !extname(path) && path[path.length - 1] !== sep) {
+    var pathEndsWithSep = path[path.length - 1] === sep
+    
+    if (err && err.code === 'ENOENT' && !extname(path) && !pathEndsWithSep) {
       // not found, check extensions
       return next(err)
     }
     if (err) return self.onStatError(err)
     if (stat.isDirectory()) return self.redirect(path)
+    if(pathEndsWithSep) return self.error(404)
+    
     self.emit('file', path, stat)
     self.send(path, stat)
   })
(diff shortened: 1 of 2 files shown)


wesleytodd:
Hey @rmhaiderali, would you be able to add some tests and docs for this?

rmhaiderali:
Hi @wesleytodd, I've included the test case, but I'm unable to come up with any documentation content since it's more of a bug fix. If you have any suggestions please let me know.

wesleytodd:
I was thinking that maybe a call out for the behavior on windows and how it differs from *nix systems would be good to document. These sort of cases require either docs or someone who develops on windows around to ensure the context is not lost, so just wanted to make sure the reason was well documented since I dont think anyone who maintains these packages develops on windows.

wesleytodd:
I am working to prepare the 1.0 release so I re-targeted this PR to that. I understand the reason you can't see a good doc to write and now that it is going out with a major it is less likely to cause churn. Would you be willing to rebase this onto 1.0? If not I can do that tomorrow before merging.

PR #233, opened 2024-07-26
fix: engines node@>=18

wesleytodd:
cc @ctcpip

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "index.js"
   ],
   "engines": {
-    "node": ">= 0.10"
+    "node": ">= 18"
   },
   "scripts": {
     "lint": "eslint .",


ctcpip (approved):

ctcpip:
I notice 0.18 is still latest in npm.  that expected?

wesleytodd:
Yes, I did that on purpose because we need to do some integration validation across all the express repos before I am comfortable with new installs getting this. I did link locally but the test were flaky (see the edit history of the PR to `express` lol) and so I was worried maybe it was something up with this, but then all the tests passed in CI and locally it seemed to have been a fluke (why I could not find) but I didn't want to go all in since this is an easy way to "soft release" the new major since I *think* it is actually solid.

PR #246, opened 2025-02-03
chore(deps): updated dependency `mime-types` to v3

Phillip9587:
This PR addresses the issue of multiple versions of the `mime-types` package being installed when using Express v5, which currently results in **two different versions** of `mime-types`. By updating the `mime-types` dependency within `send`, we can eliminate one of these redundant versions.

[Dependency Graph for Express v5](https://npmgraph.js.org/?q=express%405)


bjohansebas (approved):

bjohansebas:
In this case, I'm not sure if this update follows SemVer, as it updates multiple file extensions.

Ref: https://github.com/jshttp/mime-types/blob/master/HISTORY.md#300--2024-08-31

Phillip9587:
Closed by #251

PR #247, opened 2025-02-03
chore(deps): updated dependency `fresh` to v2

Phillip9587:
This PR addresses the issue of multiple versions of the `fresh` package being installed when using Express v5, which currently results in two different versions of `fresh`. By updating the `fresh` dependency within `send`, we can eliminate one of these redundant versions.

[Dependency Graph for Express v5](https://npmgraph.js.org/?q=express%405)


bjohansebas (approved):

UlisesGascon (approved):

wesleytodd:
https://github.com/jshttp/fresh/blob/master/HISTORY.md

We did the thing we said we wouldn't do with this one. Release *just* to drop old node versions. That said, for this case it works out because it means we can land this without any worry in a minor.

wesleytodd (approved):

PR #251, opened 2025-03-18
deps: update `mime-types` to v3.0.1

bjohansebas:
Could this be a breaking change? This might make it a breaking change in the new MIME types resolution, which is why I have doubts about it. This is also a direct dependency of Express, so I'm not sure.

Related https://github.com/expressjs/express/issues/6316


UlisesGascon (approved):
Also we need to update the `HISTORY.md` here too @bjohansebas

package.json
```diff
@@ -22,7 +22,7 @@
     "etag": "^1.8.1",
     "fresh": "^2.0.0",
     "http-errors": "^2.0.0",
-    "mime-types": "^2.1.35",
+    "mime-types": "^3.0.0",
```
bjohansebas, line 25: You're still going to release mime-types as a patch.
```suggestion
    "mime-types": "^3.0.1",
```

HISTORY.md
```diff
@@ -1,7 +1,8 @@
 unreleased
 ==================
 
-  * deps: 
+  * deps:
+    * mime-types@3.0.0
```
bjohansebas, line 5:
```suggestion
    * mime-types@3.0.1 
```

wesleytodd (approved):
Because the mime-types change was a node version drop which is already dropped here I think this is low risk from a breaking change.

PR #254, opened 2025-05-05
perf: optimize collapseLeadingSlashes function

Phillip9587:
Simplified the `collapseLeadingSlashes` function by combining the loop condition and increment into a single line, and replaced the deprecated `substr` method with the more modern `slice` method for better compatibility.


UlisesGascon (approved):

bjohansebas (approved):
I'm not sure this is a performance improvement, but I still think it's a good change.

Issue #194, opened 2020-06-05
Allow falling back to file when directory exists but doesn't have index


rchl:
It might make sense to allow falling back to files when a directory with the same name exists but doesn't have an index.html file.

(I'm using `serve-static` for unit-testing but I'm creating an issue against `send` as I think the functionality would have to be added here.)

## Reproduction steps

Options:
```
{
  extensions: ['html'],
  redirect: false,
}
```

A directory structure like so:
```
| - fr/
| - fr/foo.html
| - fr.html
| - index.html
```

Try to load `/fr` path.

404 is returned because directory with `fr` name exists but doesn't have `index.html` file.
In my case, I'd want it to fall back to the root file that actually exists and could be matched with `extensions` specified.

PR #274, opened 2025-10-01
Check extensions before performing directory redirects

dmgawel:
### Problem

When using the `extensions` option, `send` inconsistently handles path resolution depending on whether a directory exists at the requested path.

For example, given this file structure:

```
dist/
  about.html
  about/
    team.html
  contact.html
```

When requesting paths with `extensions: ['html']` configured:
- `/contact` (no directory exists) → serves `contact.html`
- `/about` (directory exists) → redirects to `/about/` instead of serving `about.html`

This inconsistency occurs because:

1. When a path doesn't exist, `send` receives an ENOENT error and tries extensions, successfully serving `contact.html`
2. When a path exists as a directory, `send` immediately calls `redirect(path)` without trying extensions first
3. The extension checking logic only runs when the initial `stat()` returns ENOENT

This behavior is problematic for static site generators and frameworks that want to serve files like `about.html` from the URL `/about` (no trailing slash) while also having subdirectories like `about/team.html`.

### Solution

Modified `sendFile()` to attempt extension resolution before redirecting when:
- The path resolves to a directory
- Extensions are configured
- The path has no file extension
- The path doesn't end with a separator

The `next()` function now tracks whether the original path was a directory via an `isDir` parameter. If all configured extensions fail to
resolve to a file and the original path was a directory, it falls back to the standard directory redirect behavior.

This ensures consistent behavior:
- `/contact` with `contact.html` present → serves `contact.html`
- `/about` with `about.html` present → serves `about.html`
- `/about` with only `about/` directory → redirects to `/about/`
- `/about` with both present → serves `about.html` (prioritizes file over directory)

### References

Resolves https://github.com/pillarjs/send/issues/194
Related to https://github.com/expressjs/serve-static/issues/138

### DIFF
--- a/index.js
+++ b/index.js
@@ -606,17 +606,27 @@ SendStream.prototype.sendFile = function sendFile (path) {
     var pathEndsWithSep = path[path.length - 1] === sep
     if (err && err.code === 'ENOENT' && !extname(path) && !pathEndsWithSep) {
       // not found, check extensions
-      return next(err)
+      return next(err, false)
     }
     if (err) return self.onStatError(err)
-    if (stat.isDirectory()) return self.redirect(path)
+    if (stat.isDirectory()) {
+      // if extensions are configured and path has no extension, try extensions first
+      if (self._extensions.length > 0 && !extname(path) && !pathEndsWithSep) {
+        return next(null, true)
+      }
+      return self.redirect(path)
+    }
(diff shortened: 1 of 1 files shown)


UlisesGascon:
Hey @dmgawel! Thanks for your PR... can you include also tests to prevent regressions in the future?

PR #293, opened 2026-01-21
chore: bring in the changes from fastify/send

### LINKED ISSUE #159: Adhere to the full Streams API
Currently `send` does not implement the Streams API (even if it inherits from `Stream`), and in fact inside `send(..).pipe(res)`, it does `stream.pipe(res)`.

Specifically, we had to use an "hack" to make it work well on Fastify. See https://github.com/fastify/fastify-static/pull/46, where we had to pipe it through a modified `PassThrough`.

I'm thinking of changing the API to something:

```js
const { stream, headers, statusCode } = await send(req, path, options)
```

As well as:

```js
send(req, path, options, (err,  { stream, headers, statusCode }) => {
  // ...
})
```

I would like to also remove the dependency on `on-finished` and some of the others modules that were needed in the 0.8/0.10 days.

I think this work should help as well with http2 compatibility as well. I'm happy to prototype this work in a forked module as well, just let me know.

cc @apapirovski

bjohansebas:
Hi! I’m bringing in the changes from fastify/send so we can work on a version that doesn’t break the current send API too much, while still getting the performance improvements from the send fork. The goal is only to bring in the existing changes for now, not to make it compatible yet. I’ve noticed that they removed some dependencies and added their own implementations of those dependencies in their fork; I think we could send those changes upstream so everyone can benefit from them. While it’s probably possible to work on an API in Node core, we also need to focus on our current issues.

base commit: https://github.com/fastify/send/commit/030392f6f8e667574e6f2cc6c1b4dba78be87529 of fastify/send

closes https://github.com/pillarjs/send/issues/159

cc: @mcollina @Uzlopak @Fdawgs @climba03003 because they are the ones who maintained the fork.
cc: @pillarjs/express-tc


package.json
```diff
@@ -6,10 +6,33 @@
   "contributors": [
     "Douglas Christopher Wilson <doug@somethingdoug.com>",
     "James Wyatt Cready <jcready@gmail.com>",
-    "Jesús Leganés Combarro <piranna@gmail.com>"
+    "Jesús Leganés Combarro <piranna@gmail.com>",
```
blakeembrey: Consistent `contributors` list.

test/send.js
```diff
@@ -24,7 +25,7 @@ var app = http.createServer(function (req, res) {
     .pipe(res)
 })
 
-describe('send(file).pipe(res)', function () {
+describe.skip('send(file).pipe(res)', function () {
```
blakeembrey: Why are we skipping these? Are they from the fork or us? Can we remove broken tests, or what is the plan?

blakeembrey (commented):
There's a few changes in the test suite from `t.` to using `node:assert` that I think we can omit from the PR and follow up on, but LGTM as a starting point to iterate from if the Fastify team is good with this.

LICENSE
```diff
@@ -1,23 +1,25 @@
-(The MIT License)
+MIT License
 
 Copyright (c) 2012 TJ Holowaychuk
 Copyright (c) 2014-2022 Douglas Christopher Wilson
+Copyright (c) 2023-present The Fastify team
```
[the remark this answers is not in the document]
ctcpip: per the guidance:

> By using this format, the project avoids having to deal with names of copyright holders, years or ranges of years, and variations on the (c) symbol.

I think the changes needs to be:

```
Copyright The Fastify Contributors.
Copyright The Express Contributors.
```

and avoid any other changes

blakeembrey:
> Yes, but that was done to maintain compatibility with Node.js 18

Is node 18 a requirement? I didn’t see that mentioned, I thought I this PR was a direct port and we’d do follow up PRs for changes.

blakeembrey (approved):
cc @mcollina @Fdawgs @Uzlopak 

I'm happy if y'all are, I'll iterate and release this similar to `cookie`. I'd like to look into further perf optimizations and any potential breaking changes from this PR I can cc @Fdawgs. I don't anticipate breaking things for fun or backward compatibility, only if there's a major performance or maintenance win hidden somewhere that requires it.

bjohansebas:
cc: @mcollina

mcollina:
Should I verify if Fastify could use this as a dependency instead of our fork?

bjohansebas:
The main goal would be to reconcile the fork. As breaking changes come up, we could ping you for feedback. At the moment, the code is essentially the same as what's in your fork of `send` from about +three months ago.

If there have been any changes on your side during the past few months, I'll bring them over this week so we can get everything merged and keep moving forward with that major release.

bjohansebas:
@blakeembrey, no changes have been made to the library since I opened the PR

PR #296, opened 2026-02-23
docs: add security considerations to the `root` option

UlisesGascon:


README.md
```diff
@@ -111,6 +111,14 @@ This can also be a string accepted by the
 
 Serve files relative to `path`.
 
+**Important:** The `root` option restricts the served path through string-level
+normalization only. It does not resolve symbolic links. If the served
```
blakeembrey, line 115: This is a little confusing because in a sense, it did "resolve" symbolic links. It doesn't validate it was within `root`. You used the word resolve in following sentence too.

How about a section below about "Security considerations" or "Symlinks", alongside "Caching".

README.md
```diff
@@ -111,6 +111,14 @@ This can also be a string accepted by the
...
+directory contains symlinks, they may resolve to files outside the root.
+
+Do not serve directories where untrusted users can create files or symlinks.
+If your application accepts file uploads or operates in a shared/multi-tenant
+environment, ensure the served directory is not writable by untrusted parties.
```
blakeembrey, line 120: I need to think about the phrasing more, but my attempt:

## Symlinks

By default, symlinks within `root` are followed and sent to users. Do not create symlinks within the `root` directory to files or directories that may be sensitive.

PR #304, opened 2026-04-11
fix: use hasOwnProperty guard for opts.index

oversteps-kr:
Replace `opts.index !== undefined` with
`Object.prototype.hasOwnProperty.call(opts, 'index')` in the SendStream constructor so that a polluted `Object.prototype.index` cannot be silently inherited and used as the directory index list.

Without this guard, any application that:
  1. calls send() without an explicit index option, and
  2. is reachable by a prototype-pollution source

can have its _index overwritten with an attacker-controlled path (e.g. '../../secret.txt'), causing sendIndex() to serve arbitrary files outside the static root.

Adds a regression test that simulates the scenario and asserts that a traversal path inherited from Object.prototype is not served.


blakeembrey:
Inheriting properties from the prototype isn't an issue to be fixed, and is by design of the language even if it's not commonly utilized. Some people may use this as a feature, since there's no reason you can't create an options class and reuse it.

It may also be somewhat surprising that `index` can be set to a relative path like that, so we could disallow that in the future, but for now if someone configures their application like that we should serve it as-is.

blakeembrey:
Here's a valid example in the wild: https://github.com/expressjs/express/blob/8e022edc9185f540a3fcecaf5e56b850d919cdac/lib/response.js#L472-L473

Although it's not a good example of this particular option, it shows how an application may be using prototype inheritance today to pass through configuration to this package and expect it to work. If we applied this fix to this or other fields, downstream applications would break.

